PR #2, opened 2015-05-29
'type' in Input config may be plugin name

halaxa:
Support for input plugin name in `'type'` input configuration key for `Zend\InputFilter\Factory::createInput()`.


halaxa:
Ping @weierophinney. Is there anything else I should add to this PR? Thank you.

PR #3, opened 2015-05-29
Change to InputFilterFactory and CollectionInputFilter to support nested input-filter config

nobesnickr:
I've had some difficulty implementing InputFilterFactory via array config when needing to implement nested CollectionInputFilter; Specifically when using zf-content-validation.

My use case is such that input will have nested entities (and associated input filter configurations), and my problems arose from having to explicitly re-define/duplicate the validators because the base InputFilterPluginManager was not available downstream.

I'm suggesting the following changes that I've made to resolve this:
1. Added the InputFilterPluginManager to new InputFilterFactory instances created via the Input Filter Abstract Service Factory.
2. Added functionality to retrieve nested InputFilters on the CollectionInputFilter via the Service Manager if available (allowing for nested InputFilter configurations).

I encourage feedback and will be happy to re-work as needed as there may be a reason the InputFilterManager wasn't supplied in the factory initially.

### DIFF
--- a/src/InputFilterAbstractServiceFactory.php
+++ b/src/InputFilterAbstractServiceFactory.php
@@ -81,6 +81,8 @@ class InputFilterAbstractServiceFactory implements AbstractFactoryInterface
             ->getDefaultValidatorChain()
             ->setPluginManager($this->getValidatorPluginManager($services));
 
+        $this->factory->setInputFilterManager($services->get('InputFilterManager'));
+
         return $this->factory;
     }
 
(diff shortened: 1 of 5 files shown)


manuakasam:
I'm not so sure about what the use-case is here. Aren't you doing something
like this?

```php
FormFilter >
    $nestedfilter = new CollectionInputFilter();
    $nestedFilter->setInputFilter(new SingleModelFilter($dependencies));

    $this->add($nestedFilter, 'my-nested-models');
```

Instead of `new SingleModelFilter()` you could also simply get access to
the pluginmanager there and get it from there (hence, even better inject
the singleModelfilter as dependency in your formFilter).

Phillip McCaffrey <notifications@github.com> schrieb am Sa., 30. Mai 2015
um 00:54 Uhr:

weierophinney:
Merged to develop for release with 2.7.0.

PR #7, opened 2015-07-23
Problems working with allowEmpty & required

weierophinney:
Hey there,

I'm trying to get my head around using Zend\InputFilter for validation. I don't use the library with Zend\Form, I try to use it for API backend data validation and filtering. In this setup we have POST data coming from outside and I would like to validate this against business rules.

Let's say, I have the attribute `foo`, which is required and allow_empty is false and I have `bar`, which is required, but needs allow_empty true, to be able to set an empty string as value. Doing it this way, isValid returns false, if I do not provide any `foo` key inside the array I provide with setData call. This is what I expect, everything finet.

But if I provided a valid `foo` attribute and I do not provide any `bar` key inside the data array, the isValid call will return true. What I would expect: failing like the first test, because the key `bar` is completly missing.

I do know, that there were many changes around isRequire and allowEmpty in last two releases. Since then we're struggeling with our defined InputFilters and trying to understand, what changes happend and how the logic behind this is thought to work now. I already have read, that the attribute continueIfEmpty is also having a role inside the InputFilter process and it is working for me, that validation rules get processed if I set it to true. 

Digging through the isValid process I recognized, that firing an isValid call is first running through $this->getRawValues(); - after that point, $data array contains `bar` with NULL value, because the data keys are generated out of defined inputFilter. So setting continueIfEmpty to true and adding an own validator to check for existing key inside my data array does not work, because it _does_ exist inside there, being populated through getRawValues call.

So, is there any way to validate provided input for existiting keys with InpuFilter? Or how can I achieve having a fixed requirement of data keys, that need to be provided, or input gets marked as invalid? I would really to avoid extra checks like

    if (!isset($postValues['bar'])) {
        throw new \InvalidArgumentException('Please provide a value for "bar"');
    }

My tests to show, what my problem is.

    class AllowEmptyTest extends \PHPUnit_Framework_TestCase
    {
        public function testInputFilterBeingRequiredAndDataNotProvided()
        {
            $inputFilter = new \Zend\InputFilter\InputFilter();
            $inputFilter->add([
                'name' => 'foo',
                'required' => true,
                'allow_empty' => false,
            ]);
            $inputFilter->add([
                'name' => 'bar',
                'required' => true,
                'allow_empty' => true,
            ]);
    
            $data = [
                'foo2' => 'xyz',
                'bar' => ''
            ];
    
            $inputFilter->setData($data);
    
            $this->assertFalse($inputFilter->isValid());
        }
    
        public function testInputFilterBeingRequiredAndAllowEmptyTrue()
        {
            $inputFilter = new \Zend\InputFilter\InputFilter();
            $inputFilter->add([
                'name' => 'foo',
                'required' => true,
                'allow_empty' => false,
            ]);
            $inputFilter->add([
                'name' => 'bar',
                'required' => true,
                'allow_empty' => true,
            ]);
    
            $data = [
                'foo' => 'xyz'
            ];
    
            $inputFilter->setData($data);
    
            $this->assertFalse($inputFilter->isValid());
        }
    }

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -210,8 +210,7 @@ class BaseInputFilter implements
      */
     public function isValid($context = null)
     {
-        $data = $this->getRawValues();
-        if (null === $data) {
+        if (null === $this->data) {
             throw new Exception\RuntimeException(sprintf(
                 '%s: no data present to validate!',
                 __METHOD__
@@ -219,7 +218,7 @@ class BaseInputFilter implements
         }
 
         $inputs = $this->validationGroup ?: array_keys($this->inputs);
-        return $this->validateInputs($inputs, $data, $context);
+        return $this->validateInputs($inputs, $this->data, $context);
     }
 
     /**
@@ -244,7 +243,24 @@ class BaseInputFilter implements
         foreach ($inputs as $name) {
             $input      = $this->inputs[$name];
 
-            // make sure we have a value (empty) for validation of context
+            // If the value is required, but not present in the data set,
+            // validation fails.
+            if (!array_key_exists($name, $data)
+                && $input instanceof InputInterface
+                && $input->isRequired()
+            ) {
+                $input->setErrorMessage('Value is required');
+                $this->invalidInputs[$name] = $input;
+
+                if ($input->breakOnFailure()) {
+                    return false;
+                }
+
+                $valid = false;
+                continue;
+            }
+
+            // Make sure we have a value (empty) for validation of context
             if (!array_key_exists($name, $data)) {
                 $data[$name] = null;
             }
--- a/test/BaseInputFilterTest.php
+++ b/test/BaseInputFilterTest.php
@@ -641,8 +641,10 @@ class BaseInputFilterTest extends TestCase
     /**
      * @dataProvider contextDataProvider()
      */
+    // @codingStandardsIgnoreStart
     public function testValidationMarksInputValidWhenAllowEmptyFlagIsTrueAndContinueIfEmptyIsTrueAndContextValidatesEmptyField($allowEmpty, $blankIsValid, $valid)
     {
+        // @codingStandardsIgnoreEnd
         $filter = new InputFilter();
 
         $data = [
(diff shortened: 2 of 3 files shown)


src/BaseInputFilter.php
```diff
@@ -219,7 +218,7 @@ public function isValid($context = null)
         }
 
         $inputs = $this->validationGroup ?: array_keys($this->inputs);
-        return $this->validateInputs($inputs, $data, $context);
+        return $this->validateInputs($inputs, $this->data, $context);
```
weierophinney: This turned out to be the key. Using `getRawValues()` was a poor solution, as it pulls values from the inputs themselves; since these return null when no value is set, there's no clean way to determine if the field was missing in the original data set!

Since the `$data` set is what was originall passed, and it's only used for its keys in `validateInputs()`, this provided the key to fixing the issue.

src/BaseInputFilter.php
```diff
@@ -244,7 +243,24 @@ protected function validateInputs(array $inputs, array $data = [], $context = nu
         foreach ($inputs as $name) {
             $input      = $this->inputs[$name];
 
-            // make sure we have a value (empty) for validation of context
+            // If the value is required, but not present in the data set,
+            // validation fails.
+            if (!array_key_exists($name, $data)
+                && $input instanceof InputInterface
+                && $input->isRequired()
+            ) {
```
weierophinney: This is the second part of the solution: if the key for the input is not present, and that input is required, then we should have a validation failure. This ensures that happens.

test/BaseInputFilterTest.php
```diff
@@ -964,4 +966,58 @@ public function testAllowEmptyTestsFilteredValueAndContinuesIfEmpty()
 
         $this->assertFalse($filter->isValid());
     }
+
+    /**
+     * @group 7
+     */
+    public function testMissingRequiredAllowedEmptyValueShouldMarkInputFilterInvalid()
```
weierophinney: This and the next test case are based on the original issue.

test/FactoryTest.php
```diff
@@ -655,15 +655,15 @@ public function testSuggestedTypeMayBePluginNameInInputFilterPluginManager()
         $pluginManager->setService('bar', new Input('bar'));
         $factory->setInputFilterManager($pluginManager);
 
-        $input = $factory->createInput(array(
+        $input = $factory->createInput([
```
weierophinney: Changes in this class are due to php-cs-fixer complaining about lack of short array syntax, and were necessary to make the PR pass CI.

weierophinney:
I think that we should likely do an LTS release with this fix as well, as the problem and solution fit a number of reports I've seen since the 2.4 release. I'll schedule that for Monday (27 July 2015).

src/BaseInputFilter.php
Mischosch: `getRawValues` usage was exactly what I spotted to be the problem for my usecase and what I didn't understood on the validation workflow side (why don't they rely on given data and take the input filter as default set¿). I thought it was intended to go this way (always having a null default), so I didn't thought of it being a "poor solution". I just always thought that I didn't found the missing puzzle piece to understand it :)

Mischosch:
Wow, @weierophinney - you're making me so happy. Thanks a lot for your done work. Next time I will try to add a possible solution as PR to my initial request. I think in this case you're def having a better overview of changes affecting other users, the history of the code and the definition of what these parameters should do or not do. (Hopefully my next problem is not inside such a sensible topic like validation.) 

I think this is a good thing for the InputFilter library. Makes it much more usable for API validation now and we do not need to refactor our code to another validation library.  Thanks again for your time and work! :fireworks:

akrabat:
>I think that we should likely do an LTS release with this fix as well, as the problem and solution fit a number of reports I've seen since the 2.4 release. I'll schedule that for Monday (27 July 2015).

I agree.

weierophinney:
Released with:

- zend-inputfilter 2.5.2
- zend-inputfilter 2.4.5
- zf2 2.4.5

(For ZF2 2.5 versions, just do a `composer update`; you'll get the new component version.)

Ocramius:
This PR broke BC for our project when upgrading from `2.4.4` -> `2.4.5`. Still investigating, but there is indeed a problem with it.

I suggest caution when upgrading until we actually find the regression (or verify that it was already fixed in `develop`)

/cc @asgrim

asgrim:
Yes indeed, this change breaks our tests when upgrading, but it seems we are relying on incorrect behaviour to run tests.

Despite reading through and agreeing with this particular change (which makes sense), it's still a BC break because previously behaviour was undefined, and projects (such as ours) are relying on this behaviour. I also understand that effort is being put in to clearly define expected behaviour now:

> ...which led to some developers and projects depending on edge cases that were not intended, and which we cannot support long-term due to the fact that they cause conflicts...

which it seems we're in this boat ;)

We're going to have to lock our composer.json to `2.4.4` until we can now get time to go through and fix our project's broken tests.

Maks3w:
#74 already claims this introduce a BC Break.

PR #10, opened 2015-07-30
InputFilter not populating/using FallbackValue for Input

weierophinney:
After updating from Version 2.3.5 to the newest Version i'm running into following error.

I'm creating an new InputFilter creating the Inputs in __construct() like this:

    class ListDataInputFilter extends InputFilter {
        public function __construct() {
            $page = new \Zend\InputFilter\Input('page');
            $page->setFallbackValue(1);
            $this->add($page);
        }
    }

Then i'm calling

    // empty to demonstrate behaviour
    $filterParams = array();
    $filter = new ListDataFilter();
    $filter->setData($filterParams);
    if(!$filter->isValid()) {
        var_dump($filter->getMessages());
    }

With Version 2.3.5 it's worked as expected. Populating the Fallback Value and not throwing an Error. But with 2.5.* instead of filling the Page Input with the Fallback Value it's empty and the Filter is not valid.

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -241,13 +241,15 @@ class BaseInputFilter implements
         $valid               = true;
 
         foreach ($inputs as $name) {
-            $input      = $this->inputs[$name];
+            $input       = $this->inputs[$name];
+            $hasFallback = ($input instanceof Input && $input->hasFallback());
 
-            // If the value is required, but not present in the data set,
-            // validation fails.
+            // If the value is required, not present in the data set, and
+            // has no fallback, validation fails.
             if (!array_key_exists($name, $data)
                 && $input instanceof InputInterface
                 && $input->isRequired()
+                && !$hasFallback
             ) {
                 $input->setErrorMessage('Value is required');
                 $this->invalidInputs[$name] = $input;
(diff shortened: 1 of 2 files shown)


weierophinney:
@disasterdrop I've verified the issue, and:

- created unit tests to cover the fallback value behavior, as we did not have any in the input filter tests (only the input tests);
- created a patch to address the issue.

Please check the tests to make sure they're testing the behavior you need, and we'll get a release out with the fixes.

weierophinney:
When approved, we'll also release a 2.4 LTS version with the fix.

akrabat:
I've reviewed and it looks good to me. Feel free to merge.

weierophinney:
@disasterdrop Why did you close this? We haven't merged the patch yet!

disasterdrop:
@weierophinney  
Sorry for that... 
I'm relative new to git and thought i would approve that your code was the right behaviour, after i testet your fix, liked you asked me to do.

src/BaseInputFilter.php
```diff
@@ -260,6 +262,18 @@ protected function validateInputs(array $inputs, array $data = [], $context = nu
...
+            // If the value is required, not present in the data set, and
+            // has a fallback, validation passes, and we set the input
+            // value to the fallback.
+            if (!array_key_exists($name, $data)
+                && $input instanceof InputInterface
+                && $input->isRequired()
+                && $hasFallback
+            ) {
+                $input->setValue($input->getFallbackValue());
+                continue;
```
Maks3w: Should the input to be append to validInputs collection?

PR #15, opened 2015-08-11
Argument 2 passed to Zend\InputFilter\BaseInputFilter::validateInputs() must be of the type array, object given

weierophinney:
This issue was introduced between version 2.5.1 and 2.5.3, i've switched back and forth between the two versions and I keep getting this error with 2.5.3 but not with 2.5.1
this happens when i try to validate a simple post data

Catchable fatal error: Argument 2 passed to Zend\InputFilter\BaseInputFilter::validateInputs() must be of the type array, object given, called in C:\crazylister\vendor\zendframework\zend-inputfilter\src\BaseInputFilter.php on line 221 and defined in C:\crazylister\vendor\zendframework\zend-inputfilter\src\BaseInputFilter.php on line 232

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -224,12 +224,12 @@ class BaseInputFilter implements
     /**
      * Validate a set of inputs against the current data
      *
-     * @param  array      $inputs
-     * @param  array      $data
+     * @param  array $inputs
+     * @param  array|ArrayAccess $data
      * @param  mixed|null $context
      * @return bool
      */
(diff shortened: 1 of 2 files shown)


weierophinney:
I've labelled as "BC Break" to indicate that the bug is *drawing attention to an existing BC Break*. The patch that I've now attached to this issue fixes it.

weierophinney:
Will release with versions 2.4.7 (LTS compatibility) and 2.5.4 (5.5+ compat).

PR #18, opened 2015-08-16
Deprecate BaseInputFilter and fix for BaseInputFilter::add() contract violation

Maks3w:
InputFlterInterface::add() specify array (input specification format) as an allowed type for $input. BaseInputFilter::add() throws exception for this case.

This commit promote the logic present in InputFilter to BaseInputFilter so BaseInputFilter becomes compatible with Interface contract.

Due InputFilter allow Traversable object, and Factory (underlying logic) allows InputProviderInterface objects then I decided promote all this types to the Interface too.

With this change BaseInputFilter is equal to InputFilter so BaseInputFilter has been declared as deprecated while InputFilter is preserved because his name is more relevant.

All of this is for to preserve the backward compatibility.


Maks3w:
For v3 `array`, `Traversable` and `InputProviderInterface` should be removed from the interface contract and promove use Factory previously.

Martin-P:
Please mark as a `BC break`, because you are changing and moving class methods and therefore possibly breaking custom code which extends `BaseInputFilter`. Same as the other PR: this is not a `bug`, merely a possible `enhancement`.

Maks3w:
@Martin-P Again. There is no BC Break. This PR fix BaseInputFilter::add does not allow input specifications (array) as the interface defines.

weierophinney:
I totally disagree with this.

The reason `BaseInputFilter` exists is so that if developers do not wish to use the *factory* capabilities, which have the overhead of plugin managers, they can. `InputFilter` extends `BaseInputFilter` in order to provide the factory capabilities specifically.

Maks3w:
I've created #31 for remove array from interface contract (which could be see as a BC break)

PR #22, opened 2015-08-20
Missing required field with fallback does not return input in getValidInput

Maks3w:
Test method says `testMissingRequiredThatAllowsEmptyWithFallbackShouldMarkInputValid` but don't assert input is returned in `getValidInput`

Should the input to be returned in `getValidInput` collection?

IMO it should because has a valid status.

Test was introduced in #10

### DIFF
--- a/test/BaseInputFilterTest.php
+++ b/test/BaseInputFilterTest.php
@@ -1069,6 +1069,8 @@ class BaseInputFilterTest extends TestCase
         $data = $filter->getValues();
         $this->assertArrayHasKey('bar', $data);
         $this->assertEquals($bar->getFallbackValue(), $data['bar']);
+        $this->assertArrayHasKey('bar', $filter->getValidInput());
+        $this->assertArrayNotHasKey('bar', $filter->getInvalidInput());
     }
 
     /**
@@ -1096,6 +1098,8 @@ class BaseInputFilterTest extends TestCase
         $data = $filter->getValues();
         $this->assertArrayHasKey('bar', $data);
         $this->assertEquals($bar->getFallbackValue(), $data['bar']);
+        $this->assertArrayHasKey('bar', $filter->getValidInput());
+        $this->assertArrayNotHasKey('bar', $filter->getInvalidInput());
     }
 
     /**


weierophinney:
If it was not present in the submitted data, and is considered optional, it should not be in the returned values, *unless it has a fallback value*. In the case of the "valid inputs" set, even though it is considered *valid*, it was never actually *validated* as it was not submitted.

Further, changing the behavior now could lead to a new category of upgrade issues, as developers have been currently working on the assumption that the valid input set will not return inputs that were not submitted.

weierophinney:
Evidently, what I stated above was the actual intent of the PR.

test/BaseInputFilterTest.php
```diff
@@ -1069,6 +1069,8 @@ public function testMissingRequiredThatAllowsEmptyWithFallbackShouldMarkInputVal
         $data = $filter->getValues();
         $this->assertArrayHasKey('bar', $data);
         $this->assertEquals($bar->getFallbackValue(), $data['bar']);
+        $this->assertArrayHasKey('bar', $filter->getValidInput());
+        $this->assertArrayNotHasKey('bar', $filter->getInvalidInput());
```
weierophinney: *both* of these should be asserting that the key is not present.

test/BaseInputFilterTest.php
```diff
@@ -1096,6 +1098,8 @@ public function testEmptyRequiredValueWithFallbackShouldMarkInputValid()
...
```
weierophinney: *both* of these should be asserting that the key is not present.

weierophinney: Actually I was wrong above; these assertions are correct, as the value *was* provided in the data set.

weierophinney:
I've pulled this locally and made the change as noted above (in the first test, indicating that the input is in neither the list of valid or invalid inputs); all tests pass. I'll merge with that change.

test/BaseInputFilterTest.php
Maks3w: But `bar` has a fallback value. If not then `assertTrue(isValid)` should be false

PR #24, opened 2015-08-20
Fix loop validation input context is mutable

### LINKED ISSUE #16: BC in validators context value
Before 2.5.2 context values was all inputs added to inputfilter but 2.5.2 changed this and now returns raw values passed to inputFilter. I Think this should be and is correct behavior, but unfortunately  now it is BC. This should be at least mentioned in release notes or fixed.

example:
```php
<?php

require __DIR__ . '/vendor/autoload.php';

class MyFilter extends Zend\InputFilter\InputFilter
{
    public function init()
    {
        $this->add([
            'name' => 'field1',
            'required' => true,
        ]);

        $this->add([
            'name' => 'field2',
            'required' => true,
            'validators' => [
                [
                    'name' => 'Zend\Validator\Callback',
                    'options'  => [
                        'callback' => function ($value, $context) {
                            var_dump($context);
                            return true;
                        },
                    ],
                ],
            ],
        ]);
    }
}

$app = Zend\Mvc\Application::init([
    'modules' => [
    ],

    'module_listener_options' => [
        'module_paths' => [
        ],

        'config_glob_paths' => [
        ],
    ],

    'input_filters' => [
        'invokables' => [
            'MyFilter' => 'MyFilter',
        ],
    ],
]);

$manager = $app->getServiceManager()->get('InputFilterManager');
$filter = $manager->get('MyFilter');

echo "\n";
echo "#------------------------------\n";
echo "# Test case\n";
echo "#------------------------------\n";
echo "\n";
$filter->setData([
    'field2' => 'foo',
]);

$filter->isValid();
```
output with versions =<2.5.1
```php
#------------------------------
# Test case
#------------------------------

array(2) {
  'field1' =>
  NULL
  'field2' =>
  string(3) "foo"
}
```

output with versions =>2.5.2
```php
#------------------------------
# Test case
#------------------------------

array(1) {
  'field2' =>
  string(3) "foo"
}
```

Maks3w:
Given two input fields the context used in each one will be different. https://github.com/zendframework/zend-inputfilter/pull/24/files#diff-2cd0c1ba642dc3e6de9ca4e2af0f5e71R290

This means inputs need to be order sensitive

```php
$inputs = ['A', 'B'];
$data = [ /* empty */ ];

$inputA->isValid($context = ['A' => null]);
$inputB->isValid($context = ['A' => null, 'B' => null]);
```

```php
// Inverse order will change $context too.
$inputs = ['B', 'A'];
$data = [ /* empty */ ];

$inputB->isValid($context = ['B' => null]);
$inputA->isValid($context = ['B' => null, 'A' => null]);
```

This PR provides fix and enhance this feature by building a default input context using the default input values when no custom context is provided.

Fix #16

Supersede and close #30


Maks3w:
Note tests are not failling just the coverage due the different weights given to each line after removal.

svycka:
looks good
also have to note that this adds something new
```php
$inputs = ['A', 'B'];
$data = ['C' => 'something'];
```
In context now we have:
```php
['A' => null, 'B' => null, 'C' => 'something']
```
Before `'C' => 'something'` would not be added to context.
But I think this is not a bug but good feature. Maybe add test to keep it :)

Maks3w:
Yep I opt for use default input raw value for build the context (usually null, sometimes mixed). Test it's placed for this enhancement.

weierophinney:
As I noted on #25, I'm a bit worried that changing test assumptions is indicative of a BC break; also, like #25, the logic looks reasonable. I'll do a more thorough review early next week.

src/BaseInputFilter.php
```diff
@@ -236,6 +236,8 @@ protected function validateInputs(array $inputs, $data = [], $context = null)
             $data = $this->getRawValues();
         }
 
+        $inputContext = $context ?: (array_merge($this->getRawValues(), (array) $data));
```
Maks3w: May this should use the filtered version instead the raw value

weierophinney: Considering that the original from two lines above was using the raw values, we should keep it that way for now. We can address again with the next major version.

weierophinney:
@Maks3w Can you rebase this, please? When I merged, there was a conflict, and my attempts to resolve the conflict have left me with failing tests.

src/BaseInputFilter.php
Maks3w: Note before the context was always null for optional values. getRawValues already change the behavior if user call setValue before

Maks3w:
Since #25 optional fields is not call what was expected. I've to rewrite the test

Maks3w:
Done. Now depends from #30

src/ArrayInput.php
```diff
@@ -50,6 +60,20 @@ public function getValue()
...
+
+        if (! $hasValue && $hasFallback) {
+            $this->setValue($this->getFallbackValue());
+            return true;
+        }
+
+        if (! $hasValue && $required) {
+            $this->setErrorMessage('Value is required');
+            return false;
+        }
```
weierophinney: This looks good per our discussion last week on IRC.

For those reading:

- If the value was not present but has a fallback, we set it to the default value and consider it valid.
- If the value was not present, but required, we mark it as invalid.

src/ArrayInput.php
```diff
@@ -64,9 +88,9 @@ public function isValid($context = null)
             }
             $result = $validator->isValid($value, $context);
             if (!$result) {
-                if ($this->hasFallback()) {
+                if ($hasFallback) {
                     $this->setValue($this->getFallbackValue());
-                    $result = true;
+                    return true;
```
weierophinney: This is the third part of the fallback value series:

- If the value is invalid, but has a fallback value, we reset the value to the fallback, and mark the value valid.

src/ArrayInput.php
```diff
@@ -50,6 +60,20 @@ public function getValue()
...
+        $required = $this->isRequired();
+        $hasFallback = $this->hasFallback();
+
+        if (! $hasValue && $hasFallback) {
+            $this->setValue($this->getFallbackValue());
+            return true;
+        }
+
+        if (! $hasValue && $required) {
+            $this->setErrorMessage('Value is required');
```
weierophinney: Does this resolve the issue with the NotEmpty messages that have occurred since 2.4? It seems like we should set it in the same format, using the same key, in order to retain the previous behavior?

Issue #14, opened 2015-08-04
NotEmpty validator doesn't override the required attribute


larsnystrom:
A `NotEmpty` validator should override the `required` check on an input, but in some cases it doesn't. I'm sorry I haven't been able to pin point the exact location of the bug, but here's the code to reproduce it. In test case 1 both `NotEmpty` validators overrides the `required` attribute, which can be seen by the error messages. In test case 2 however, `field2` doesn't use it's `NotEmpty` validator but instead falls back to it's internal check to see if the value is empty.

    <?php

    require __DIR__ . '/vendor/autoload.php';

    class MyFilter extends Zend\InputFilter\InputFilter
    {
        public function init()
        {
            $this->add([
                'name' => 'field1',
                'required' => true,
                'validators' => [
                    [
                        'name' => 'Zend\Validator\NotEmpty',
                        'break_chain_on_failure' => true,
                        'options' => [
                            'messages' => [
                                'isEmpty' => "Field1 cannot be empty",
                            ],
                        ],
                    ],
                ],
            ]);

            $this->add([
                'name' => 'field2',
                'required' => true,
                'validators' => [
                    [
                        'name' => 'Zend\Validator\NotEmpty',
                        'break_chain_on_failure' => true,
                        'options' => [
                            'messages' => [
                                'isEmpty' => "Field2 cannot be empty",
                            ],
                        ],
                    ],
                ],
            ]);
        }
    }

    $app = Zend\Mvc\Application::init([
        'modules' => [
        ],

        'module_listener_options' => [
            'module_paths' => [
            ],

            'config_glob_paths' => [
            ],
        ],

        'input_filters' => [
            'invokables' => [
                'MyFilter' => 'MyFilter',
            ],
        ],
    ]);

    $manager = $app->getServiceManager()->get('InputFilterManager');
    $filter = $manager->get('MyFilter');

    echo "\n";
    echo "#------------------------------\n";
    echo "# Test case 1\n";
    echo "#------------------------------\n";
    echo "\n";
    $filter->setData([
    ]);

    if (!$filter->isValid()) {
        var_dump($filter->getMessages());
    }

    echo "\n";
    echo "#------------------------------\n";
    echo "# Test case 2\n";
    echo "#------------------------------\n";
    echo "\n";
    $filter->setData([
        'field1' => [
            'test',
        ],
    ]);

    if (!$filter->isValid()) {
        var_dump($filter->getMessages());
    }

The output I get is:

    #------------------------------
[...]

PR #25, opened 2015-08-20
Fix missing optional fields should not be validated

Maks3w:
Prevent the assumption of `null` value to be a synonym of value not set. Null values are valid input values (like JSON object with null values)

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -256,6 +256,7 @@ class BaseInputFilter implements
 
             $hasFallback = ($input instanceof Input && $input->hasFallback());
 
+            // If input is optional (not required) and value is not set then ignore.
             if (!array_key_exists($name, $data)
                 && !$input->isRequired()
             ) {
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -231,11 +231,6 @@ class BaseInputFilter implements
      */
     protected function validateInputs(array $inputs, $data = [], $context = null)
     {
-        // backwards compatibility
-        if (empty($data)) {
-            $data = $this->getRawValues();
-        }
-
         $this->validInputs   = [];
         $this->invalidInputs = [];
         $valid               = true;
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -261,6 +261,12 @@ class BaseInputFilter implements
 
             $hasFallback = ($input instanceof Input && $input->hasFallback());
 
+            if (!array_key_exists($name, $data)
+                && !$input->isRequired()
+            ) {
+                continue;
+            }
+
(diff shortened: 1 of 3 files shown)


Maks3w:
Most basic test does not work due https://github.com/zendframework/zf2/pull/5271/ `getRawValues()` returns a list of pairs `Input => null` which make this logic to fail.

svycka:
I think this explains and fixes https://github.com/zendframework/zf2/issues/6638

Maks3w:
It's not related. This does not change the meaning of any empty value. This only treat the case when the value is not set (not exists !== empty).

svycka:
There was impossible to check if value exists or not, so I used notEmpty validator to cover this, but that was not ideal because there was some values treated as empty anyway. This fixes my issue because now I can check not set, is set or not empty, before this PR that was impossible. As I understand this PR adds this functionality.

weierophinney:
I'm concerned that this changes test assumptions, as that generally indicates a BC break. Looking at the logic, however, it appears sound. I'll review more fully early next week.

test/BaseInputFilterTest.php
```diff
@@ -508,21 +510,32 @@ public function testValidationSkipsFieldsMarkedNotRequiredWhenNoDataPresent()
...
+        $optionalInput->method('getName')
+            ->willReturn($optionalInputName)
+        ;
+        $optionalInput->expects($this->never())
+            ->method('isValid')
+        ;
+        $data = [];
 
-        $bar = new Input();
-        $bar->getValidatorChain()->attach(new Validator\Digits());
```
Maks3w: Input::isValid() `if ($empty && ! $required && ! $continueIfEmpty) { return true; }`

Make this to pass. Test refactor generalize this and make the test resilient to InputInterface implementations with the use of the mock.

src/BaseInputFilter.php
```diff
@@ -231,11 +231,6 @@ public function isValid($context = null)
      */
     protected function validateInputs(array $inputs, $data = [], $context = null)
     {
-        // backwards compatibility
-        if (empty($data)) {
-            $data = $this->getRawValues();
```
Maks3w: Removal  of this make consistent the output of #14.

Basically this BC does not ensure the same data when the original is empty. 

Due the fact by default all fields will have a valid value (null) It's impossible detect when the field was really set.

Maks3w: Basically this should have been removed in #7 due the same arguments described in this comment https://github.com/zendframework/zend-inputfilter/pull/7/files#r35374042

Maks3w:
Note: It's impossible to undo all BC Breaks introduced in ZF 2.4. The split of `null` between not set and valid value can't be undone.

Issue #12, opened 2015-08-03
Rename `continue_if_empty` to `validate_if_empty` and default to true


larsnystrom:
The configuration parameter `continue_if_empty` in `Zend\InputFilter\Input` doesn't convey much meaning. Continue doing what? The name has no apparent link to the validation process, where it is an essential parameter. 

I also think it should default to `true`. Because of #11 I marked a field with `required == false` and added a `NotEmpty` validator instead, and later found out that the validator didn't run. What's the point of a `NotEmpty` validator if it doesn't run when the value is empty?

weierophinney:
The idea behind `continue_if_empty` was "continue validating even if the value is empty." I'm in agreement with you that renaming it might make it more clear. However, doing so would also be a substantial BC break.

My suggestion is that if we do this, we do it as follows:

- Next minor release:
  - Rename `continueIfEmpty` property to `validateIfEmpty`.
  - Introduce `setValidateIfEmpty()` method; it will manipulate the `validateIfEmpty` flag.
  - Introduce `validateIfEmpty()` method; it will return the `validateIfEmpty` flag.
  - Rewrite `setContinueIfEmpty()` to proxy to `setValidateIfEmpty()` and raise an `E_USER_DEPRECATED` error indicating the latter method should be used instead.
  - Rewrite `continueIfEmpty()` to proxy to `validateIfEmpty()` and raise an `E_USER_DEPRECATED` error indicating the latter method should be used instead.
  - Update `Zend\InputFilter\Factory` to:
    - on detection of `continue_if_empty` in a specification, raise an `E_USER_DEPRECATED` error indicating that `validate_if_empty` should be used instead; have it pass the value to `setValidateIfEmpty()`.
    - detect `validate_if_empty` in a specification and pass it to `setValidateIfEmpty()`.
  - Update an logic internally that uses the `*continueIfEmpty` methods to use the new methods.
- Next major release:
  - Remove the deprecated methods and support for the deprecated key.

Any takers?

weierophinney:
I'm noting this as a BC Break. Even if we follow the above, by marking a method as deprecated, we *will* break some applications. (In particular, I know that when Apigility users have display_errors enabled, this breaks the payloads returned.) However, it *is* BC-safe, as the original methods are preserved, and a migration path is offered which will remove the errors.

larsnystrom:
That seems reasonable. What do you think of the change of default value?

weierophinney:
[...]

PR #26, opened 2015-08-20
Deprecate EmptyContextInterface, InputInterface::setAllowEmpty & allowEmpty

Maks3w:
Deprecate magic logic for auto attach a NonEmpty validator with breakChainOnFailure = true

Instead append NonEmpty validator when desired.

```php
$input = new Zend\InputFilter\Input();
$input->setContinueIfEmpty(true);
$input->setAllowEmpty(true);
$input->getValidatorChain()->attach(new Zend\Validator\NotEmpty(), /* break chain on failure */ true);
```

Close https://github.com/zendframework/zend-inputfilter/issues/12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,6 +121,32 @@ All notable changes to this project will be documented in this file, in reverse 
   (previously, it would consider the data set valid, and auto-initialize the
   missing input to `null`).
 
+## 2.4.8 - TBD
+
+### Added
+
+- Nothing.
+
+### Deprecated
+
+- [#26](https://github.com/zendframework/zend-inputfilter/pull/26) Deprecate magic logic for auto attach a NonEmpty
+ validator with breakChainOnFailure = true. Instead append NonEmpty validator when desired.
+
+  ```php
+  $input = new Zend\InputFilter\Input();
+  $input->setContinueIfEmpty(true);
+  $input->setAllowEmpty(true);
+  $input->getValidatorChain()->attach(new Zend\Validator\NotEmpty(), /* break chain on failure */ true);
+  ```
+
+### Removed
+
+- Nothing.
+
+### Fixed
+
+- Nothing.
+
 ## 2.4.7 - 2015-08-11
 
 ### Added
--- a/src/EmptyContextInterface.php
+++ b/src/EmptyContextInterface.php
@@ -9,15 +9,22 @@
 
 namespace Zend\InputFilter;
 
+/**
+ * @deprecated 2.4.8 Add Zend\Validator\NotEmpty validator to the ValidatorChain.
+ */
 interface EmptyContextInterface
 {
     /**
+     * @deprecated 2.4.8 Add Zend\Validator\NotEmpty validator to the ValidatorChain and set this to `true`.
+     *
      * @param bool $continueIfEmpty
      * @return self
      */
     public function setContinueIfEmpty($continueIfEmpty);
 
     /**
+     * @deprecated 2.4.8 Add Zend\Validator\NotEmpty validator to the ValidatorChain. Should always return `true`.
+     *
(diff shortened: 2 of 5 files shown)


Maks3w:
Since ZF 2.4 there are lot of bugs and headaches about the `Required VS AllowEmpty VS ContinueIfEmpty matrix`

This simplify the problem by removing the last two and forcing the user to attach the `NotEmpty` validator when he/she wants that behavior.

src/EmptyContextInterface.php
```diff
@@ -9,15 +9,22 @@
 
 namespace Zend\InputFilter;
 
+/**
+ * @deprecated 2.5.5 Add Zend\Validator\NotEmpty validator to the ValidatorChain.
+ */
 interface EmptyContextInterface
 {
     /**
+     * @deprecated 2.5.5 Add Zend\Validator\NotEmpty validator to the ValidatorChain.
```
Maks3w: May we should describe `setContinueIfEmpty(true)` should be do too.

src/InputInterface.php
```diff
@@ -15,6 +15,8 @@
 interface InputInterface
 {
     /**
+     * @deprecated 2.5.5 Add Zend\Validator\NotEmpty validator to the ValidatorChain.
```
Maks3w: May we should describe `setAllowEmpty(true)` should be do too.

src/EmptyContextInterface.php
```diff
@@ -9,15 +9,22 @@
 
 namespace Zend\InputFilter;
 
+/**
+ * @deprecated 2.5.5 Add Zend\Validator\NotEmpty validator to the ValidatorChain.
```
Maks3w: Do we should deprecate this starting from 2.4.8?

weierophinney:
Any delegation needs to happen in add new minor version; as such, it needs to be targeted at 2.6.0.

Maks3w:
I propose to do this on next patch because I found this a thing to be done ASAP. Anyway the most important thing is if we all agree this should be removed

weierophinney:
Any deprecations MUST happen at a new minor release. After we get the other issues ironed out, we can do a 2.6.0 release with these changes.

Maks3w:
Rebased on top of develop

Maks3w:
Anyway I think deprecations should be backport to the lowest version for early adoptions of workarounds.

As soon as the developer avoid deprecated features as soon he/she will upgrade to major versions with less effort.

Maks3w:
I think we have two classes of deprecatations.

* New features which make obsolete old one. This should be deprecated in a new minor because includes a new feature.
* Removal of code without new feature. This could be deprecated at any moment because there is no new code related to it.

CHANGELOG.md
```diff
@@ -2,6 +2,32 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
+## 2.6.0 - TBD
```
weierophinney: I'm confused at this point. In IRC, you're indicating this should be merged against the LTS release as well; to my mind, that's saying it should be part of the 2.4 and 2.5 series, and not require a new version. If this is truly backwards compatible, I'm inclined to do so, as it gives the cleanest path forwards for existing users to upgrade.

However, here, you're indicating the next new minor version, 2.6.0. I'm not sure if this is due to the feedback I provided previously, or if this was what you originally intended. However, it's contrary to being submitted against the `master` branch, which targets 2.5.

So, confusion all around on my part.

Can you clarify, please?

Maks3w: > @weierophinney said: Any delegation needs to happen in add new minor version; as such, it needs to be targeted at 2.6.0.

Maks3w:
Ready for 2.4.8 LTS

PR #27, opened 2015-08-20
Remove test introduced on zendframework/zf2#3608

Maks3w:
Seems only test validators are called. Lot tests already do that.

See zendframework/zf2#3608


weierophinney:
This is testing a specific edge-case, specifically that if an array value is passed to a standard input (vs an ArrayInput), validation can still work. As such, removal of the test is likely to re-introduce a bug that was resolved previously.

weierophinney:
I'd accept a change in the test name to make that more clear. :)

Maks3w:
I'll remove this in another PR along the rest of test enhancements I'm doing.

PR #30, opened 2015-08-21
Delegate to InputInterface::isValid when data not exists

Maks3w:
This change is BC compatible with #10. Both depends of the specific subclass `Input`

About #7 there is things for to discuss if this is BC break or not.

Basically the question to discuss it's if `InputInterface::isValid()` should be the responsible to decide all cases where an input is invalid or if we want `BaseInputFilter` to decide that. Note I used here `BaseInputFilter` and not `InputFilterInterface` because only that class had so many matches about the `InputInterface|Input` state.

Revert interpretation of #22 change on merge

Supersede and close #29


Maks3w:
At this moment this pending of discussion.

TODO remove tests from BaseInputFilter and place it on Input

weierophinney:
This looks sane; please rebase and proceed, and let me know when you're ready for final review.

Maks3w:
@weierophinney This become broken due the assumption of fallback values for missing data does not add the input to the valid inputs collection.

Maks3w:
@weierophinney This undo the change on #22 while merging. Fallback value shouldn't never to be exposed and used for build validation rules.

test/BaseInputFilterTest.php
```diff
@@ -1036,86 +1036,6 @@ public function testEmptyValuePassedForRequiredButAllowedEmptyInputShouldMarkInp
...
-            'foo' => 'xyz',
-            'bar' => null,
-        ]);
-        $this->assertTrue($filter->isValid(), 'Empty input with fallback value should mark input filter as valid');
-        $data = $filter->getValues();
-        $this->assertArrayHasKey('bar', $data);
-        $this->assertEquals($bar->getFallbackValue(), $data['bar']);
-        $this->assertArrayHasKey('bar', $filter->getValidInput());
-        $this->assertArrayNotHasKey('bar', $filter->getInvalidInput());
-    }
```
weierophinney: Do the above tests fail with the change? Or are equivalent tests introduced? If these tests still pass, I'd like to keep them in place, as they document changes that were introduced with #10.

Maks3w: There is no test ensuring any valid input (i.e isValid > true) must appear on getValidInput result.

This will be added on the test refactor where all outputs are tested at once for ensure there is no edge cases.

Maks3w: And yep, test was failing https://travis-ci.org/zendframework/zend-inputfilter/jobs/77456631#L258

This Pr undo #22

PR #31, opened 2015-08-21
Remove array type for input on InputFilterInterface::add()

Maks3w:
array capabilities only exists for InputFilter class but not for their parent BaseInputFilter which is responsible of implement the interface.

People with tools warning about unsafe type should typehint against Zend\InputFilter\InputFilter instead Zend\InputFilter\InputFilterInterface

Reverts:

* zendframework/zf2#2304
* 8fa473579e3cd3ef24765f00bfeb2af152a0ea2a
* 93b9bd0538ce223a1808dd4d03d2d23bc776ac8b

### DIFF
--- a/src/InputFilterInterface.php
+++ b/src/InputFilterInterface.php
@@ -19,7 +19,7 @@ interface InputFilterInterface extends Countable
     /**
      * Add an input to the input filter
      *
-     * @param  InputInterface|InputFilterInterface|array $input
+     * @param  InputInterface|InputFilterInterface $input
      * @param  null|string $name Name used to retrieve this input
      * @return InputFilterInterface
      */


Maks3w:
IMO InputFilter::add() method should be deprecated and encourage the use of external call to Factory for achieve the same functionallity.

weierophinney:
The one problem I see with this is that removing it from the interface would mean that `InputFilter`'s implementation automatically violates the interface. While I totally understand your view point that we may want to recommend using an external factory *instead of* `add()`+array specification, that doesn't address the *current* and *existing* behavior.

weierophinney:
A better solution would be to specify in the interface that the implementation MUST handle one or more of the argument types specified, and CAN raise an exception for any it cannot process. This would make it valid for all implementations shipped.

weierophinney:
Updated with suggestions on merge.

PR #33, opened 2015-08-21
Test suite improvements

Maks3w:
From the first commit up to `Improve PHPDoc types` (a6e9b0c) It's basically CS.

To be backported to 2.4 LTS:

* `Consolidation of Factory Exception expectations + namespe bug`
* `[Factory] Fix unsafe call non InputInterface methods`
* `Consolidation of InputInterface::merge and Fix unsafe merging`
* `Fix allow_empty / continue_if_empty cases`

This PR could be partial merge at least up to the commit `Assert objects are instances of the expected type` and the three commits from above

PHP < 5.5 Incompatibilties:

* Use of FQDN::class constant instead FQDN string
* Use of yield keyword

### DIFF
--- a/test/FileInputTest.php
+++ b/test/FileInputTest.php
@@ -19,21 +19,13 @@ use Zend\Validator;
  */
 class FileInputTest extends InputTest
 {
-    /** @var FileInput */
-    protected $input;
-
-    public function setUp()
-    {
-        $this->input = new FileInput('foo');
-        // Upload validator does not work in CLI test environment, disable
-        $this->input->setAutoPrependUploadValidator(false);
-    }
-
     public function testValueMayBeInjected()
     {
+        $input = $this->createDefaultInput();
+
         $value = ['tmp_name' => 'bar'];
-        $this->input->setValue($value);
-        $this->assertEquals($value, $this->input->getValue());
+        $input->setValue($value);
+        $this->assertEquals($value, $input->getValue());
     }
 
     public function testRetrievingValueFiltersTheValue()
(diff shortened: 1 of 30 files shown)


test/FileInputTest.php
```diff
@@ -9,38 +9,215 @@
...
+        ];
+        // @formatter:on
+
+        // Expand data template matrix for each possible input value.
+        // Description => [$allowEmpty, $isRequired, $isValid, $value, $expectedIsValid, $expectedMessages]
+        foreach ($dataTemplates as $dataTemplateDescription => $dataTemplate) {
+            $temporalTemplate = $dataTemplate;
+            foreach ($dataTemplate[4] as $valueDescription => $value) {
+                $temporalTemplate[4] = current($value);
+                yield $dataTemplateDescription . ' / ' . $valueDescription => $temporalTemplate;
```
weierophinney: One note: if we want to backport this to 2.4, we'll need to switch from using generators to aggregating.

Maks3w:
I've open an issue on zend-validator for make 0 and 0.0 to be non empty https://github.com/zendframework/zend-validator/pull/35

test/InputTest.php
```diff
@@ -9,95 +9,207 @@
...
+            if ($cIETrueRule[3] === false) { // ValidChain: false
+                $cIETrueRule[6] = array_merge($cIETrueRule[6], $validatorMessages);  // $expectedMessages
+            }
+            $inputValidationRules[$cIETrueDescription] = $cIETrueRule;
+
+            $cIEFalseDescription = 'CIEmpty: F, ' . $ruleDescription;
+            $cIEFalseRule = $rule;
+
+            array_unshift($cIEFalseRule, !$continueIfEmpty);
+            // ContinueIfEmpty = false && AllowEmpty = true && valueIsEmpty
```
Maks3w: This is not very pretty. This loop becomes too complex

Maks3w:
I think "required: false" should not be an exclusion for not reject the input is "allow_empty:false"

PR #40, opened 2015-08-29
Remove test about Input setters permutation

Maks3w:
These tests asserts the state of Input is always the same not matter the order when invoke the setters.

Sometime ago there was specific tweaks where setterA change the default value for propertieB. This feature was removed so this tests are verifyng "Feature X" not longer exists.

This given me to think if we should to test every permutation on setters call and I found it's not. There is no reason for to test a feature not longer exists.

PD: setRequired is already verified by testRequiredFlagIsMutable

### DIFF
--- a/test/InputTest.php
+++ b/test/InputTest.php
@@ -411,34 +411,6 @@ class InputTest extends TestCase
         $this->assertTrue($input2->allowEmpty());
     }
 
-    /**
-     * @group 7445
-     */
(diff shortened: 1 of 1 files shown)


Maks3w:
Ping @weierophinney

Maks3w:
Extra: If someone want to test the setters permutations then there is no need to call isValid, just query the object under test and verify it has the desired state (i.e. setA setB setC > assert GetA,GetB GetC)

PR #41, opened 2015-08-30
Refactor tests for group 7448 as a data set matrix

Maks3w:
* Refactor tests as a data set matrix which is more visual

* Fix a case where empty values in ArrayInput does not work in the same way as Input does when is not required

This development is on top of #40


test/InputTest.php
```diff
@@ -412,446 +321,138 @@ public function testMergeRetainsAllowEmptyFlag()
...
-            'required-first-empty' => [$requiredFirst, ''],
-            'required-last-empty'  => [$requiredLast, ''],
-            'required-first-array' => [$requiredFirst, []],
-            'required-last-array'  => [$requiredLast, []],
+        // @codingStandardsIgnoreStart
+        $dataTemplates=[
+            // Description => [$isRequired, $allowEmpty, $continueIfEmpty, $validator, [$values], $expectedIsValid, $expectedMessages]
+            'Required: T; AEmpty: T; CIEmpty: T; Validator: T' => [ $isRequired,  $aEmpty,  $cIEmpty, $validatorValid  , $emptyValues,  $isValid, []],
+            'Required: T; AEmpty: T; CIEmpty: T; Validator: F' => [ $isRequired,  $aEmpty,  $cIEmpty, $validatorInvalid, $emptyValues, !$isValid, $validatorMsg],
+            'Required: T; AEmpty: T; CIEmpty: F; Validator: X' => [ $isRequired,  $aEmpty, !$cIEmpty, $validatorNotCall, $emptyValues,  $isValid, []],
```
Maks3w: Makes sense to have `allow_empty:true` and `continue_if_empty:false`?

No validator will be invoked ever.

weierophinney: Yes, it makes sense. Consider the following scenario:

- Address 2 is required, and must be submitted.
- **An empty Address 2 is acceptable**
- But a non-empty Address 2 needs to pass certain validation criteria.

That's the sort of scenario that uses those three criteria with those three values.

weierophinney: > Will be ever true no matter what value was set.

Not true. *If a value is present, it will be passed to attached validators.* So, the set of valid values are:

- Empty values
- Non-empty values that pass the validation chain.

Maks3w: Non-Empty Address2 never hit the validator due `Continue:False`

See the test has an assertion for ensure is never call

PR #42, opened 2015-09-01
Consolidate InputFilterPluginManager tests

Maks3w:


weierophinney:
Just a note: while this will be backported for 2.4 it will require some substantial massaging to work on PHP 5.3. I *have* verified that 4.5.0, at the least, still supports PHP 5.3, which should make the bulk of the work translating short array notation to full array notation, and expanding `::class` constants.

PR #43, opened 2015-09-01
Consolidate Factory tests + Fixes

Maks3w:
* Fix unsafe call non InputInterface methods
* Exceptions was thrown as `Zend\Filter\Exception` instead `Zend\InputFilter\Exception`

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -10,7 +10,6 @@
 namespace Zend\InputFilter;
 
 use Traversable;
-use Zend\Filter\Exception;
 use Zend\Filter\FilterChain;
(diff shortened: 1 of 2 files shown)


src/Factory.php
```diff
@@ -220,12 +219,26 @@ public function createInput($inputSpecification)
                     }
                     break;
                 case 'continue_if_empty':
+                    if (!$input instanceof Input) {
+                        throw new Exception\RuntimeException(sprintf(
+                            '%s "continue_if_empty" can only set to inputs of type "%s"',
+                            __METHOD__,
+                            Input::class
+                        ));
+                    }
```
weierophinney: I like this; makes it far safer.

PR #46, opened 2015-09-02
Expand test matrix with nonempty value scenarios

Maks3w:


Maks3w:
image

Note: 0.0 should it's already fixed https://github.com/zendframework/zend-validator/pull/35

test/InputTest.php
```diff
@@ -495,18 +497,37 @@ public function isRequiredVsAllowEmptyVsContinueIfEmptyVsIsValidProvider()
         // @codingStandardsIgnoreStart
         $dataTemplates=[
             // Description => [$isRequired, $allowEmpty, $continueIfEmpty, $validator, [$values], $expectedIsValid, $expectedMessages]
-            'Required: T; AEmpty: T; CIEmpty: T; Validator: T' => [ $isRequired,  $aEmpty,  $cIEmpty, $validatorValid  , $emptyValues,  $isValid, []],
-            'Required: T; AEmpty: T; CIEmpty: T; Validator: F' => [ $isRequired,  $aEmpty,  $cIEmpty, $validatorInvalid, $emptyValues, !$isValid, $validatorMsg],
-            'Required: T; AEmpty: T; CIEmpty: F; Validator: X' => [ $isRequired,  $aEmpty, !$cIEmpty, $validatorNotCall, $emptyValues,  $isValid, []],
-            'Required: T; AEmpty: F; CIEmpty: T; Validator: T' => [ $isRequired, !$aEmpty,  $cIEmpty, $validatorValid  , $emptyValues,  $isValid, []],
```
Maks3w: This was weird. `Allow empty: false`, `value: empty` => `isValid` ?

svycka: the whole thing about required, allow_empty and continue_if_empty is weird.. I think they all should be removed and required behavior should be changed  from notEmpty to is present. Then would be possible add NotEmpty just when it is required. because InputFilter is used a lot with json API and this would make more sense. But Zend\Form could add NotEmpty when it needs. I understand that this would be big BC break but now there is to many problems and questions about those three :)

svycka: unfortunately I don't have photo of my colleague when I showed him these tests :)

Maks3w: @svycka The feature is declared deprecated since yesterday. Unfortunatelly this is part of the LTS version and have to be maintained.

Maks3w: @svycka You have another try for to capture the photo. Tell him this array does not represent the real test cases. The below loop multiply the number of cases for each possible real value (not groups) IIRC the number of cases generated > 90

svycka: oh. Good to know this :)

PR #55, opened 2015-09-02
Make symmetric the scenarios when required is true/false

Maks3w:
There is no reason for to have differences when required is true or false and the value is always set.

### DIFF
--- a/src/ArrayInput.php
+++ b/src/ArrayInput.php
@@ -84,10 +84,6 @@ class ArrayInput extends Input
         $result    = true;
         foreach ($values as $value) {
             $empty = ($value === null || $value === '' || $value === []);
-            if ($empty && !$this->isRequired() && !$this->continueIfEmpty()) {
-                $result = true;
-                continue;
-            }
             if ($empty && $this->allowEmpty() && !$this->continueIfEmpty()) {
                 $result = true;
                 continue;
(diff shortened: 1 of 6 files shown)


Maks3w:
@weierophinney I've reopen this because restore the behavior of 2.3.

Maks3w:
From 2.3.9 I can upgrade without BC Breaks except for this one and the validator renaming for PHP7 compliance (Int > IsInt)

Maks3w:
Workaround (Recommended, rely on `allow_empty:false/continue_if_empty:false` is now deprecated):

Before:

```php
array(
    'name' => 'password',
    'required' => false,
    'filters' => array(),
    'validators' => array(),
    'error_message' => 'This field cannot be empty',
    'allow_empty' => true,
    'continue_if_empty' => true,
)
```

After:
```php
array(
    'name' => 'password',
    'required' => false,
    'filters' => array(),
    'validators' => array(
        0 => array(
            'break_chain_on_failure' => true,
            'name' => 'Zend\\Validator\\NotEmpty',
            'options' => array(),
        ),
    ),
    'error_message' => 'This field cannot be empty',
    'allow_empty' => false,
    'continue_if_empty' => false,
)
```

Maks3w:
We can again discard this because the feature to fix is now deprecated and exists workaround

Issue #28, opened 2015-08-21
'Value is required' message should be localizable


tflin:
We had the issue that 'Value is required' message can not be localized to Chinese as it is hardcoded in the framework code. The custom-message NotEmpty validator cannot override this error message in our case. Please help to make the hardcoded message localizable.

Maks3w:
> @weierophinney said: Does this resolve the issue with the NotEmpty messages that have occurred since 2.4? It seems like we should set it in the same format, using the same key, in order to retain the previous behavior?

From #24

weierophinney:
Fixed with #61

stefanotorresi:
@weierophinney i'm afraid it actually isn't, see #67

PR #60, opened 2015-09-03
Not set default required error message always

Maks3w:
Skip if a previous one was set

Partial fix for #28.

This does not allow to translate automatically with multiple localizations.

### DIFF
--- a/test/InputTest.php
+++ b/test/InputTest.php
@@ -175,6 +175,19 @@ class InputTest extends TestCase
         $this->assertEquals(['Value is required'], $input->getMessages(), 'getMessages() value not match');
     }
 
+    public function testRequiredWithoutFallbackAndValueNotSetThenFailWithCustomErrorMessage()
+    {
+        $input = $this->input;
+        $input->setRequired(true);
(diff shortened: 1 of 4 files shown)


test/InputTest.php
```diff
@@ -175,6 +175,19 @@ public function testRequiredWithoutFallbackAndValueNotSetThenFail()
         $this->assertEquals(['Value is required'], $input->getMessages(), 'getMessages() value not match');
     }
 
+    public function testRequiredWithoutFallbackAndValueNotSetThenFailWithCustomErrorMessage()
+    {
+        $input = $this->input;
+        $input->setRequired(true);
+        $input->setErrorMessage('fooErrorMessage');
```
weierophinney: This isn't what #28 was getting at.

Prior to 2.4, a required input that was provided an empty value would return `[ 'isEmpty' => "Value is required and can't be empty"]` for the messages; that's what the reporter wants to happen again.

I'll work on this using your patch as a start, and submit an alternate PR.

weierophinney:
Closing in favor of #61.

PR #61, opened 2015-09-03
Provide NotEmpty::IS_EMPTY validation message for required input

weierophinney:
Per #28, since 2.4, we've not been returning error messages in the expected format when a value is required but no value (or an empty value) was provided for the input.

This patch builds on #60, and update the test expectations to those reported in #28.

### DIFF
--- a/test/InputTest.php
+++ b/test/InputTest.php
@@ -11,7 +11,6 @@ namespace ZendTest\InputFilter;
 
 use PHPUnit_Framework_MockObject_MockObject as MockObject;
 use PHPUnit_Framework_TestCase as TestCase;
-use ReflectionProperty;
 use stdClass;
 use Zend\Filter;
 use Zend\Filter\FilterChain;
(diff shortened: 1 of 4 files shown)


test/InputTest.php
```diff
@@ -175,6 +177,35 @@ public function testRequiredWithoutFallbackAndValueNotSetThenFail()
...
+        );
+        $messages = $input->getMessages();
+        $this->assertInternalType('array', $messages);
+        $this->assertArrayHasKey(NotEmpty::IS_EMPTY, $messages);
+
+        $notEmpty = new NotEmpty();
+        $r = new ReflectionProperty($notEmpty, 'messageTemplates');
+        $r->setAccessible(true);
+        $messageTemplates = $r->getValue($notEmpty);
+        $this->assertEquals(
```
Maks3w: What about `assertAttributeEquals`

test/InputTest.php
```diff
@@ -175,6 +177,35 @@ public function testRequiredWithoutFallbackAndValueNotSetThenFail()
...
+        $input->setRequired(true);
+
+        $this->assertFalse(
+            $input->isValid(),
+            'isValid() should always return false when no fallback value is present, '
+            . 'the input is required, and no data is set.'
+        );
+        $messages = $input->getMessages();
+        $this->assertInternalType('array', $messages);
+        $this->assertArrayHasKey(NotEmpty::IS_EMPTY, $messages);
```
Maks3w: I prefer compare against the whole `getMessages` return. By this way we are sure any other message appears

Maks3w:
Following the reasoning of to restore the `NonEmpty` behavior the correct thing could be inject NonEmpty validator instead `setErrorMessage`

weierophinney:
> Following the reasoning of to restore the NonEmpty behavior the correct thing could be inject NonEmpty validator instead setErrorMessage

I consider this a stop-gap, as we'll be removing it (as we've deprecated this use case with earlier patches). As such, we're just trying to mimic previous behavior at this point.

Maks3w:
This part is not deprecated (still smelling of new) so this could be reused in future developments.

The API suggested should be reused on v3.

I found this way does not allow customization through input specifications.

PR #63, opened 2015-09-03
Ensure custom error messages are used for required missing inputs

weierophinney:
This builds on #61 to add a few more test cases covering when required inputs have custom error messages set.

The code already correctly handles it, but we did not have test cases for the scenarios previously.

### DIFF
--- a/test/InputTest.php
+++ b/test/InputTest.php
@@ -191,6 +191,19 @@ class InputTest extends TestCase
         $this->assertRequiredValidationErrorMessage($input);
     }
 
+    public function testRequiredWithoutFallbackAndValueNotSetThenFailReturnsCustomErrorMessageWhenSet()
+    {
+        $input = $this->input;
+        $input->setRequired(true);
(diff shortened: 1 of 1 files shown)


test/InputTest.php
```diff
@@ -208,6 +221,24 @@ public function testRequiredWithoutFallbackAndValueNotSetProvidesNotEmptyValidat
         $this->assertRequiredValidationErrorMessage($input);
     }
 
+    /**
+     * @group 28
+     * @group 60
+     */
+    public function testRequiredWithoutFallbackAndValueNotSetProvidesCustomErrorMessageWhenSet()
```
Maks3w: I don't see the difference with testRequiredWithoutFallbackAndValueNotSetThenFailReturnsCustomErrorMessageWhenSet

Both have the same body. The only difference is the PHPUnit error message for assertFalse

weierophinney:
As @Maks3w pointed out: these were the tests I rewrote from #60 - we needed both sets, after all!

PR #72, opened 2015-10-06
BaseInputFilter::populate() calls resetValue() for missing arrayinputs

BreiteSeite:
`hasValue` returns true for ArrayInputs whether or not a value was provided via `InputFilter::setData()`. This can't be the expected behavior of this function.

However, i had to refactor the existing test `BaseInputFilterTest::testPopulateSupportsArrayInputEvenIfDataMissing`. I don't know the implications regarding BC because neither the docblock nor the commit message provided a useful use-case for this test-method.

Feedback is very welcome.

Thank you for considering my pull request.

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -513,7 +513,7 @@ class BaseInputFilter implements
                 }
 
                 if ($input instanceof ArrayInput) {
-                    $input->setValue([]);
+                    $input->resetValue();
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


src/BaseInputFilter.php
```diff
@@ -513,7 +513,7 @@ protected function populate()
                 }
 
                 if ($input instanceof ArrayInput) {
-                    $input->setValue([]);
+                    $input->resetValue();
```
Maks3w: The whole if can be removed then because ArrayInput is a child of Input

BreiteSeite: Good catch, thank you.

Maks3w:
Merged for next bugfix. I don't consider this need to be backported to LTS

/cc  @weierophinney

PR #73, opened 2015-10-07
Hotfix/69

### LINKED ISSUE #69: Input::prepareRequiredValidationFailureMessage return value
After upgrading to 2.5.5,  Apigility validators are not showing expected error message for required fields if error_message is not set. Previously, it was showing the error message from a NotEmpty validator if present in the validator chain. 
I think the newly introduced prepareRequiredValidationFailureMessage() method should return existing NotEmpty validator message, if present in the validator chain.

larsnystrom:
This fixes #69 and also touches #28. Basically it uses the message of an existing `NotEmpty` validator instead of creating a new `NotEmpty` validator when a required input is missing.

### DIFF
--- a/src/Input.php
+++ b/src/Input.php
@@ -491,7 +491,20 @@ class Input implements
      */
     protected function prepareRequiredValidationFailureMessage()
     {
-        $notEmpty = new NotEmpty();
+        $chain = $this->getValidatorChain();
+        $validators = $chain->getValidators();
+
+        foreach ($validators as $validator) {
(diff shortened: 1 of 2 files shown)


akrabat:
👍 This would solve a problem I currently have…

Maks3w:
Please take a look to #62

larsnystrom:
I think we should merge this and #67 (and make them compatible, since they overlap a little). #62 seems more like a refactor which will cause BC breaks. I'm not against that, but it won't happen until 3.0 or 4.0, while this can be merged without any BC breaks right now.

PR #76, opened 2015-10-12
not empty validator trasnlation

svycka:


svycka:
did not sow that method is protected :(

PR #86, opened 2016-01-08
[WIP] Migrate to ZF3 components

gianarb:
ping @ezimuel 
Work in progress

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,12 +14,12 @@
     },
     "require": {
         "php": ">=5.5",
-        "zendframework/zend-filter": "~2.5",
-        "zendframework/zend-validator": "^2.5.3",
-        "zendframework/zend-stdlib": "~2.5"
+        "zendframework/zend-filter": "dev-develop as 2.6.0",
+        "zendframework/zend-validator": "dev-develop",
+        "zendframework/zend-stdlib": "dev-develop as 2.8.0"
     },
     "require-dev": {
-        "zendframework/zend-servicemanager": "~2.5",
+        "zendframework/zend-servicemanager": "dev-develop as 2.7.0",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "^4.5"
     },
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -10,6 +10,7 @@
 namespace Zend\InputFilter;
 
 use Traversable;
+use Zend\ServiceManager\ServiceManager;
 use Zend\Filter\FilterChain;
 use Zend\ServiceManager\ServiceLocatorInterface;
 use Zend\Stdlib\ArrayUtils;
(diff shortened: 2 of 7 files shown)


src/Factory.php
```diff
@@ -117,7 +117,7 @@ public function clearDefaultValidatorChain()
     public function setInputFilterManager(InputFilterPluginManager $inputFilterManager)
     {
         $this->inputFilterManager = $inputFilterManager;
-        $serviceLocator = $this->inputFilterManager->getServiceLocator();
+        $serviceLocator = $this->inputFilterManager;
         if ($serviceLocator && $serviceLocator instanceof ServiceLocatorInterface) {
```
Maks3w: Is this field still required?

Maks3w:
Since this PR is similar to the rest of SMv3 PRs I think we should release a beta version of SMv3.

Following SMv3 betas should be backward compatible.

/cc @weierophinney

weierophinney:
I plan to tag v2.7.0 and v3 of the SM early next week, add we've finished
ironing out the migration issues. @ezimuel and I will then with up a plan
to update the various components to target v2.7 (which will allow usage in
both v2 and v3).

Maks3w:
Sincerely I don't like nothing raise dependencies from 2.5 to 2.7 like has been do with stdlib.

1. Teorically should work with 2.5
2. If we have introcuced by accident any bug or backward incompatibility change then we will have a new nightmare like the changes was done between 2.3 and 2.4

src/InputFilterPluginManager.php
```diff
@@ -26,11 +28,16 @@ class InputFilterPluginManager extends AbstractPluginManager
      *
      * @var string[]
      */
-    protected $invokableClasses = [
+    protected $aliases = [
         'inputfilter' => InputFilter::class,
         'collection'  => CollectionInputFilter::class,
     ];
 
+    protected $factories = [
```
Maks3w: Please add a docblock or inheritdoc

src/InputFilterPluginManager.php
```diff
@@ -43,7 +50,11 @@ class InputFilterPluginManager extends AbstractPluginManager
      */
     public function __construct(ConfigInterface $configuration = null)
     {
-        parent::__construct($configuration);
+        if (null === $configuration) {
+            $configuration = [];
+        }
+
+        parent::__construct(new ServiceManager(), $configuration);
```
Maks3w: I think this dependency has been exposed in the rest of SMv3 PRs.

https://github.com/zendframework/zend-validator/pull/49/files#diff-f1414faa3e531f8d828ad37211d2dd28R89

composer.json
```diff
@@ -15,17 +15,23 @@
     "require": {
         "php": ">=5.5",
         "zendframework/zend-filter": "~2.5",
-        "zendframework/zend-validator": "^2.5.3",
-        "zendframework/zend-stdlib": "~2.5"
+        "zendframework/zend-validator": "dev-feature/new-service-manager",
```
Maks3w: V3 support exists on develop branch. Does not need of this. Also we could wait to the next week for use an stable release

composer.json
```diff
@@ -15,17 +15,23 @@
...
+        "zendframework/zend-stdlib": "dev-develop as 2.8.0"
```
Maks3w: Please undo this. What feature exists only in 2.8?

composer.json
```diff
@@ -15,17 +15,23 @@
...
     "require-dev": {
-        "zendframework/zend-servicemanager": "~2.5",
+        "zendframework/zend-servicemanager": "dev-develop as 2.7.0",
         "fabpot/php-cs-fixer": "1.7.*",
         "phpunit/PHPUnit": "^4.5"
     },
     "suggest": {
         "zendframework/zend-servicemanager": "To support plugin manager support"
     },
+    "repositories": [
```
Maks3w: Remove this

composer.json
weierophinney: The SM updates for that component exist on the develop branch.

As you have not been involved on the other component migrations, I respectfully request that you refrain from further review of this PR; @ezimuel and I will complete this review, as we've been involved with all other component updates to date, and know the changes required intimately.

composer.json
gianarb: I am working with the new version of zend-validator atm

ezimuel: @Maks3w this is a WIP so we should wait for comments until the author complete the PR, thanks. /cc @gianarb

composer.json
Maks3w: @weierophinney After SMv3 release composer should pick the most recent stdlib version automatically without any explicit requirement. This is important because if SM has a problem with the stdlib version then should be fixes on SM composer and don't in each component, application, ...

kynx:
I *think* your max nesting error is caused by https://github.com/zendframework/zend-inputfilter/blob/develop/src/Factory.php#L316 - calling get() is returning an existing instance. The following does the trick for me:

```
$inputFilter = $this->getInputFilterManager()->build($type);
```

kynx:
Scratch that - `protected $shareByDefault = false;` -> `protected $sharedByDefault = false;` in `InputFilterManager` is the correct thing to do methinks.

PR #95, opened 2016-02-13
ServiceManager v2-v3 compatibility

kynx:
This builds on #86 adding ServiceManager v2/v3 compatibility.

The v2 tests will fail pending https://github.com/zendframework/zend-validator/pull/51


.travis.yml
```diff
@@ -17,9 +17,15 @@ matrix:
     - php: 5.5
       env:
         - EXECUTE_CS_CHECK=true
+    - php: 5.5
+      env:
+        - SERVICE_MANAGER_VERSION="^2.7.5"
```
Maks3w: I disagree with this. The correct is to test lowest version (composer --prefer-lowest) and newest versions. All versions released in the middle are expected to be compatible and fulfill the API contract.

When we should to test specific versions? When the code HAS an specific logic branch testing if some feature is available i.e. `(component_version >= 5.0) call new API else call old API`. And this thing should be do with only 1 PHP version and not 5.5. and 5.6 whcih make me the following question?

Why PHP 7 and HHVM are not duplicated?

I think the test matrix proposed in this PR is not necessary.

kynx: Hum, OK, didn't know about --prefer-lowest (not my preference often ;) I've just been following what I've seen @weierophinney do in PRs for zend-i18n etc.

I noticed @ezimuel has done it slightly differently, with a SERVICE_MANAGER_V2=(true|false). So something like:

```
before_install:
  - if [[ $SERVICE_MANAGER_V2 != 'true' ]]; then composer require --dev --no-update
  - if [[ $SERVICE_MANAGER_V2 == 'true' ]]; then composer require --dev --no-update --prefer-lowest
```

?

kynx: s/composer install/composer update/ - install doesn't do `--prefer-lowest`

kynx:
Nice one @Maks3w ! That makes flipping between versions when testing locally one heck of lot easier too :)

kynx:
@Maks3w : doesn't look like the `--prefer-lowest` is going to work.

The fail is due to the minimum symfony/file version specified by php-cd-fixer: 2.1. The `notPath()` method it's failing on was introduced in 2.2, and even that is pretty ancient.

weierophinney:
@kynx and @Maks3w — I tried it out with a couple of components this past week, and ran into similar issues; development tools were causing failures by composer to resolve dependencies. Specifying the upper and/or lower bounds as separate tests seems to work consistently and without issue.

Maks3w:
@weierophinney Probably is an issue with the cached dependencies. Could you detail what issues did you had so I could investigate?

Recently I have refactored a Symfony Test suite from to test every Supported version (Currenlty 4) to only to test lowest and newest with the only issue of reset the cache. 

Previous: https://travis-ci.org/Maks3w/FR3DLdapBundle/builds/108739742
Current: https://travis-ci.org/Maks3w/FR3DLdapBundle/builds/109023062

src/InputFilterPluginManager.php
```diff
@@ -22,55 +23,100 @@
...
     protected $shareByDefault = false;
 
+
     /**
-     * @param ConfigInterface $configuration
+     * @param ContainerInterface $parentLocator
+     * @param array $config
      */
-    public function __construct(ConfigInterface $configuration = null)
+    public function __construct(ContainerInterface $parentLocator, array $config = [])
```
weierophinney: First argument must also accept null for full BC with v2; I'll update that when merging. (Discovered this the hard way with zend-view 2.6.0.)

src/InputFilterPluginManager.php
```diff
@@ -82,11 +128,35 @@ public function validatePlugin($plugin)
...
+     *
+     * @param  mixed                      $plugin
+     * @return void
+     * @throws Exception\RuntimeException if invalid
+     */
+    public function validatePlugin($plugin)
+    {
+        try {
+            $this->validate($plugin);
+        } catch (InvalidServiceException $e) {
```
weierophinney: This should be `Exception\RuntimeException`, no? (I'll update on merge)

weierophinney: Never mind, was confusing this with the next line. Nothing to see here, move along...

src/InputFilterPluginManager.php
```diff
@@ -82,11 +128,35 @@ public function validatePlugin($plugin)
...
+            throw new Exception\RuntimeException($e->getMessage(), $e->getCode(), $e);
+        }
+    }
+
+    public function shareByDefault()
```
weierophinney: Why are you adding this?

src/InputFilterAbstractServiceFactory.php
```diff
@@ -47,6 +74,19 @@ public function canCreateServiceWithName(ServiceLocatorInterface $inputFilters,
...
+     * Determine if we can create a service with name (v2)
+     *
+     * @param ServiceLocatorInterface $serviceLocator
+     * @param $name
+     * @param $requestedName
+     * @return bool
+     */
+    public function canCreateServiceWithName(ServiceLocatorInterface $serviceLocator, $name, $requestedName)
+    {
+        return $this->canCreate($serviceLocator, $requestedName);
```
weierophinney: This method, when invoked, can only happen from v2, so this is where the check for the parent locator should occur (versus in `canCreate()`). (Similar should occur for `createServiceWithName()` / `__invoke()`.) I'll make that change on merge.

src/InputFilterAbstractServiceFactory.php
```diff
@@ -94,7 +128,7 @@ protected function getFilterPluginManager(ServiceLocatorInterface $services)
             return $services->get('FilterManager');
         }
 
-        return new FilterPluginManager();
+        return new FilterPluginManager(new ServiceManager());
```
weierophinney: `s/new ServiceManager()/$services/` (in both this and the `getValidatorPluginManager()` method); we already have the instance, so why create a new one? :smile:

I'll make that change on merge.

PR #98, opened 2016-03-27
Issue/21 Use plurals for valid/invalid input get in InputFilterInterface

jeanCarloMachado:


Maks3w:
Thx but this won't happen as is. 

1. This is not enough important and break the compatibility is not justfies. 
2. This change will be happen in a future refactor for stateless input where the result will have his own API

PR #110, opened 2016-06-11
Fixes issue with using InputFilterAbstractFactory with application services

weierophinney:
As reported in #123, under zend-servicemanager v2, the `InputFilterAbstractFactory` raises a fatal error when calling `getServiceLocator()` if the container passed to the abstract factory is not a plugin manager, but instead the application-level service container.

This patch updates the implementation to add some detection to determine if the container is a plugin manager when used under v2, and, if so, only then attempt to pull the parent service container. Additionally, it will now use itself if no parent container is present (this will typically cause the `canCreate()` test to fail anyways).

All typehints on `ServiceLocatorInterface` were modified to `ContainerInterface` unless they were part of an existing interface definition, to ensure inner consistency, and forwards compatibility with v3.

### DIFF
--- a/src/InputFilterAbstractServiceFactory.php
+++ b/src/InputFilterAbstractServiceFactory.php
@@ -12,6 +12,7 @@ namespace Zend\InputFilter;
 use Interop\Container\ContainerInterface;
 use Zend\Filter\FilterPluginManager;
 use Zend\ServiceManager\AbstractFactoryInterface;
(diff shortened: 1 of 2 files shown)


ppeiris:
Hi Mathew 

I still get the following error !

Fatal error: Class 'Zend\InputFiler\InputFilterAbstractServiceFactory' not found in /var/ftridp/vendor/zendframework/zend-servicemanager/src/ServiceManager.php on line 314

PR #115, opened 2016-08-18
`getUnknown` in CollectionInputFilter

webimpress:
Method `getUnknown` works wrong on `CollectionInputFilter`.
It was returning all fields, not only unknown.

It should work similar to `isValid` method - so unknown fields should be checked for each dataset and result should be collection of unknown fields (to be more specific pairs `field => value`).

So for example: we have input filter with field "foo" (and we are using this input filter in our CollectionInputFilter). The input data are:
```php
[
    ['foo' => 'bar'],
    ['baz' => 'bar'],
]
```

so the second dataset (index 1!) has unknown field `baz`.

The result of `getUnknown` method should be:
```php
[
    1 => ['baz' => 'bar'],
]
```

but currently is exactly the same as the input data.

In this PR fixed behavior of this method in `CollectionInputFilter` + tests.
Also updated method `hasUnknown` to reuse `getUnknown` results.

### DIFF
--- a/src/CollectionInputFilter.php
+++ b/src/CollectionInputFilter.php
@@ -252,4 +252,33 @@ class CollectionInputFilter extends InputFilter
     {
         return $this->collectionMessages;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getUnknown()
+    {
+        if (!is_array($this->data) || !$this->data) {
+            throw new Exception\RuntimeException(sprintf(
+                '%s: no data present!',
+                __METHOD__
+            ));
+        }
+
+        $inputFilter = $this->getInputFilter();
+
+        $unknownInputs = [];
+        foreach ($this->data as $key => $data) {
+            if (!is_array($data)) {
+                $data = [];
+            }
+            $inputFilter->setData($data);
+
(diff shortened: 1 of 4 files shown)


webimpress:
Related to https://github.com/zfcampus/zf-content-validation/issues/82 - `CollectionInputFilter::getUnknown` returns always all fields.

src/CollectionInputFilter.php
```diff
@@ -252,4 +252,33 @@ public function getMessages()
...
+                '%s: no data present!',
+                __METHOD__
+            ));
+        }
+
+        $inputFilter = $this->getInputFilter();
+
+        $unknownInputs = [];
+        foreach ($this->data as $key => $data) {
+            if (!is_array($data)) {
```
webimpress: Not entirely sure about this condition here. I've added it, because the same we have in `isValid` loop (line 175 in this file).

weierophinney: As discussed in IRC, I'll make a separate PR to validate incoming data in `CollectionInputFilter::setData()`, as both this method and `isValid()` are doing this, and, ideally, these should be considered invalid data sets.

PR #120, opened 2016-09-13
Move Coverage reporting to latest env

Koopzington:
Requiring php-coveralls results in a full update of all dependencies instead of composer installing the versions mentioned in the composer.lock


.travis.yml
```diff
@@ -42,12 +42,12 @@ matrix:
     - php: 5.6
       env:
         - DEPS=locked
-        - TEST_COVERAGE=true
         - DEPLOY_DOCS="$(if [[ $TRAVIS_BRANCH == 'master' && $TRAVIS_PULL_REQUEST == 'false' ]]; then echo -n 'true' ; else echo -n 'false' ; fi)"
         - PATH="$HOME/.local/bin:$PATH"
     - php: 5.6
       env:
         - DEPS=latest
+        - TEST_COVERAGE=true
```
localheinz: Should we collect coverage on a more recent PHP version instead, for example, PHP7?

PR #122, opened 2016-09-27
Added alias for FQCN of InputFilterPluginManager

Koopzington:


test/ConfigProviderTest.php
```diff
@@ -0,0 +1,43 @@
...
+        'factories' => [
+            InputFilterPluginManager::class => InputFilterPluginManagerFactory::class,
+        ],
+    ];
+
+    public function testProvidesExpectedConfiguration()
+    {
+        $provider = new ConfigProvider();
+        $this->assertEquals($this->config, $provider->getDependencyConfig());
+        return $provider;
```
localheinz: Can we remove this and instead just create the provider in the test below again?

Having tests depend on each other generally is not a good idea - while this is still a simple example, it makes it hard to follow. Also, one should be possible to run a single test, too.

```diff
diff --git a/test/ConfigProviderTest.php b/test/ConfigProviderTest.php
index e3a85d3..a5dbd0d 100644
--- a/test/ConfigProviderTest.php
+++ b/test/ConfigProviderTest.php
@@ -30,14 +30,17 @@ class ConfigProviderTest extends \PHPUnit_Framework_TestCase
     {
         $provider = new ConfigProvider();
         $this->assertEquals($this->config, $provider->getDependencyConfig());
-        return $provider;
     }

-    /**
-     * @depends testProvidesExpectedConfiguration
-     */
-    public function testInvocationProvidesDependencyConfiguration(ConfigProvider $provider)
+    public function testInvocationProvidesDependencyConfiguration()
     {
-        $this->assertEquals(['dependencies' => $provider->getDependencyConfig()], $provider());
+        $provider = new ConfigProvider();
+
+        $expected = [
+            'dependencies' => $provider->getDependencyConfig(),
+
+        ];
+
+        $this->assertEquals($expected, $provider());
     }
 }
```

test/ConfigProviderTest.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace ZendTest\InputFilter;
+
+use Zend\InputFilter\ConfigProvider;
+use Zend\InputFilter\InputFilterAbstractServiceFactory;
+use Zend\InputFilter\InputFilterPluginManager;
+use Zend\InputFilter\InputFilterPluginManagerFactory;
+
+class ConfigProviderTest extends \PHPUnit_Framework_TestCase
```
localheinz: Can we mark this test `final`?

test/ConfigProviderTest.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    private $config = [
```
localheinz: Since this field is only used in a single test, can we inline it?

```diff
diff --git a/test/ConfigProviderTest.php b/test/ConfigProviderTest.php
index e3a85d3..a6f13fb 100644
--- a/test/ConfigProviderTest.php
+++ b/test/ConfigProviderTest.php
@@ -14,22 +14,23 @@ use Zend\InputFilter\InputFilterPluginManagerFactory;

 class ConfigProviderTest extends \PHPUnit_Framework_TestCase
 {
-    private $config = [
-        'abstract_factories' => [
-            InputFilterAbstractServiceFactory::class,
-        ],
-        'aliases' => [
-            'InputFilterManager' => InputFilterPluginManager::class,
-        ],
-        'factories' => [
-            InputFilterPluginManager::class => InputFilterPluginManagerFactory::class,
-        ],
-    ];
-
     public function testProvidesExpectedConfiguration()
     {
         $provider = new ConfigProvider();
-        $this->assertEquals($this->config, $provider->getDependencyConfig());
+
+        $expected = [
+            'abstract_factories' => [
+                InputFilterAbstractServiceFactory::class,
+            ],
+            'aliases' => [
+                'InputFilterManager' => InputFilterPluginManager::class,
+            ],
+            'factories' => [
+                InputFilterPluginManager::class => InputFilterPluginManagerFactory::class,
+            ],
+        ];
+
+        $this->assertEquals($expected, $provider->getDependencyConfig());
         return $provider;
     }
```

weierophinney:
Thanks, @Koopzington

PR #126, opened 2016-11-13
New coding standard

Koopzington:
Usage of the newly introduced zendframework/zend-coding-standard for CS checks.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8",
-        "squizlabs/php_codesniffer": "^2.6.2",
-        "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3"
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -20,8 +20,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8",
-        "squizlabs/php_codesniffer": "^2.6.2",
-        "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3"
+        "zendframework/zend-servicemanager": "^2.7.5 || ^3.0.3",
+        "zendframework/zend-coding-standard": "~1.0.0"
```
localheinz: Are there any objections here to use the `^` operator instead?

weierophinney: Yes - new feature releases may introduce new sniffs, which could lead to test failures. Pinning it to a specific minor version prevents that, and ensures we opt in to those new sniffs deliberately.

localheinz: Ha, I see!

PR #131, opened 2017-03-06
Regression fix

marhub:
Fix to allow setting multiple validation fields in form of $name = ['field1','field2'] when those fields are simple Zend\InputFilter\Input


weierophinney:
This needs unit tests, please, to ensure a breaking change is not made again in the future.

PR #137, opened 2017-05-16
Ensure `input_filters` config is honored in non-zend-mvc contexts

weierophinney:
Per https://discourse.zendframework.com/t/validatormanager-not-calling-custom-validator-factory/109/5?u=matthew the `input_filters` config key is not honored currently unless the application is within a zend-mvc context. This is due to the fact that `Zend\InputFilter\Module` wires configuration for the `Zend\ModuleManager\Listener\ServiceListener` in order to push merged service configuration into the plugin during bootstrap; no similar logic is available when not in a zend-mvc context, however.

This patch fixes that situation by modifying the `InputFilterPluginManagerFactory` to do the following:

- If a `ServiceListener` service exists, it returns the plugin manager immediately (old behavior).
- Otherwise, it checks for the `config` service, and, if found, a `input_filters` key with an array value. When found, it feeds that value to a `Zend\ServiceManager\Config` instance and uses that to configure the plugin manager before returning it.

### DIFF
--- a/src/InputFilterPluginManagerFactory.php
+++ b/src/InputFilterPluginManagerFactory.php
@@ -8,6 +8,7 @@
 namespace Zend\InputFilter;
 
 use Interop\Container\ContainerInterface;
+use Zend\ServiceManager\Config;
 use Zend\ServiceManager\FactoryInterface;
 use Zend\ServiceManager\ServiceLocatorInterface;
 
@@ -27,7 +28,30 @@ class InputFilterPluginManagerFactory implements FactoryInterface
      */
     public function __invoke(ContainerInterface $container, $name, array $options = null)
     {
-        return new InputFilterPluginManager($container, $options ?: []);
+        $pluginManager = new InputFilterPluginManager($container, $options ?: []);
+
+        // If this is in a zend-mvc application, the ServiceListener will inject
+        // merged configuration during bootstrap.
+        if ($container->has('ServiceListener')) {
+            return $pluginManager;
+        }
+
+        // If we do not have a config service, nothing more to do
+        if (! $container->has('config')) {
+            return $pluginManager;
+        }
+
+        $config = $container->get('config');
+
+        // If we do not have input_filters configuration, nothing more to do
+        if (! isset($config['input_filters']) || ! is_array($config['input_filters'])) {
+            return $pluginManager;
+        }
+
+        // Wire service configuration for input_filters
+        (new Config($config['input_filters']))->configureServiceManager($pluginManager);
+
+        return $pluginManager;
     }
 
     /**
--- a/test/InputFilterPluginManagerFactoryTest.php
+++ b/test/InputFilterPluginManagerFactoryTest.php
@@ -81,4 +81,99 @@ class InputFilterPluginManagerFactoryTest extends TestCase
         $filters = $factory->createService($container->reveal());
         $this->assertSame($plugin, $filters->get('test'));
     }
+
+    public function testConfiguresInputFilterServicesWhenFound()
+    {
+        $inputFilter = $this->prophesize(InputFilterInterface::class)->reveal();
+        $config = [
+            'input_filters' => [
+                'aliases' => [
+                    'test' => 'test-too',
+                ],
+                'factories' => [
+                    'test-too' => function ($container) use ($inputFilter) {
+                        return $inputFilter;
+                    },
+                ],
+            ],
+        ];
+
+        $container = $this->prophesize(ServiceLocatorInterface::class);
+        $container->willImplement(ContainerInterface::class);
+
+        $container->has('ServiceListener')->willReturn(false);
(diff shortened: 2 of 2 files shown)


weierophinney:
Failing tests are against `latest` target, and are due to PHPUnit version mismatch issues; failures only affect `InputFilterPluginManagerCompatibilityTest`; I've run the new tests manually with latest dependencies, and they pass as expected.

svycka:
@weierophinney does this mean that we no longer require to write our own factory for zend-expressive? also do you plan to add this to other plugin managers like validator, filter, view, controllers and others?

froschdesign:
@svycka 
>  also do you plan to add this to other plugin managers like validator, filter, view, controllers and others?

Already done:

zendframework/zend-filter/pull/56
https://github.com/zendframework/zend-form/pull/164
https://github.com/zendframework/zend-validator/pull/168
https://github.com/zendframework/zend-log/pull/74
https://github.com/zendframework/zend-i18n/pull/74
https://github.com/zendframework/zend-hydrator/pull/59

svycka:
oh nice didn't see this

mtymek:
I would also add the factory for `Zend\InputFilter\Factory` that injects filter and validator plugin managers to appropriate chain. 

Here's how I implemented it in my library:
```php
use Interop\Container\ContainerInterface;
use Zend\Filter\FilterPluginManager;
use Zend\InputFilter\Factory;
use Zend\InputFilter\InputFilterPluginManager;
use Zend\Validator\ValidatorPluginManager;

class InputFilterFactoryFactory
{
    public function __invoke(ContainerInterface $container)
    {
        $factory = new Factory($container->get(InputFilterPluginManager::class));
        $factory->getDefaultFilterChain()->setPluginManager($container->get(FilterPluginManager::class));
        $factory->getDefaultValidatorChain()->setPluginManager($container->get(ValidatorPluginManager::class));

        return $factory;
    }
}
```

This is required to configure populate input filter using `add` method with array definition:
```php
class ContactInputFilter extends InputFilter
{
    public function __construct()
    {
        $this->add(
            [
                'name' => 'first_name',
                'filters' => [
                    ['name' => 'Foo'],
                ],
                'validators' => [
                    ['name' => 'Bar'],
                ],
            ]
        );

        $this->add(
            [
                'name' => 'content',
                'filters' => [
                    ['name' => 'StringTrim'],
                ],
            ]
        );
    }
}
```

svycka:
@mtymek are you sure this is required? I don't remember doing this. Maybe you not added filter and validators config providers to config?

froschdesign:
@mtymek 
Do you mean these code lines:

- https://github.com/zendframework/zend-inputfilter/blob/master/src/Factory.php#L45
- https://github.com/zendframework/zend-inputfilter/blob/master/src/Factory.php#L120
- https://github.com/zendframework/zend-inputfilter/blob/master/src/InputFilterPluginManager.php#L97

mtymek:
@froschdesign lines you mentioned will only work under `zend-mvc`, because FQCN is not used when pulling plugin manager for the container:

```php
$factory->getDefaultFilterChain()->setPluginManager($container->get('FilterManager'));
```

My code uses `FilterPluginManager::class` instead of just `'FilterManager'`. Do you see the nuance? ;-)

svycka:
@mtymek then I think this should be solved here https://github.com/zendframework/zend-filter/blob/master/src/ConfigProvider.php#L33 by adding FilterPluginManager::class and making 'FilterManager' as alias. This would solve this no?

froschdesign:
@mtymek 
> lines you mentioned will only work under zend-mvc

…or you have registered already these managers.

mtymek:
@svycka indeed, this could work as well. Maybe it would be enough to add the aliases to `ConfigProvider` class?

@froschdesign I don't understand your remark. Can you elaborate?

weierophinney:
@mtymek — regarding this:

> ```php
> $factory->getDefaultFilterChain()->setPluginManager($container->get(FilterPluginManager::class));
> $factory->getDefaultValidatorChain()->setPluginManager($container->get(ValidatorPluginManager::class));
> ```

I'm not sure why you need that at all. zend-filter defines the `FilterManager` service in its `ConfigProvider`, and zend-validator defines `ValidatorManager` in its own. `InputFilterPluginManager` pulls these two services when populating the `Factory` instance with the default plugin managers. Your changes above are using service names we do not define in the existing `ConfigProvider` definitions, which means they shouldn't work at all... unless your module is defining them.

I agree that we should likely add the FQCN as the service name pointing to the factory, and the shorter names as aliases to those, but after the proposed changes in this PR and the new tags I've created on the other components, it should all work seamlessly.

mtymek:
Right :) The reason you see it is that is that I working on my lib before ConfigProvider was available in `zend-inputfilter`. Sorry for adding more confusion!

weierophinney:
@mtymek No worries! Glad it's all sorted!

PR #138, opened 2017-05-21
Missing `false` as generic empty value

autowp:


Erikvv:
I don't think this is useful to consider `false` as empty, I don't remember ever wanting that. 

But it is consistent with Zend\Validator\NotEmpty

autowp:
Usage example is posting json to RESTful API and filter incoming data

Erikvv:
I don't understand what you're trying to say.

Ocramius:
Tests are required anyway: a test will clarify the ambiguity

weierophinney:
I agree with @Erikvv here. If you are submitting values via an API, and have an actual boolean `false` value submitted, the value _is not empty_. In those cases, you're likely expecting a boolean, and you would want to validate it as such. If the value is supposed to be another type entirely, I'd argue allowing `false` as an empty value would be inconsistent, because if you `allow_empty`, you're now getting a value of the wrong type.

As such, I'm closing this request. If you feel strongly about it, re-open with unit tests, and discussion of the specific use cases.

Issue #119, opened 2016-08-24
Class Zend\ServiceManager\AbstractPluginManager not found


mkroeders:
When you run `zendframework/zend-inputfilter` as a stand alone library you get `Class Zend\ServiceManager\AbstractPluginManager not found` with the following code

```
$filter = new \Zend\InputFilter\InputFilter();
$input = new \Zend\InputFilter\Input('test');
$filter->add($input);

$filter->setData([]);

$filter->isValid();
$filter->getMessages();
```

This is due to the fact of `\Zend\InputFilter\Input` line 500, where the plugin chain is used. Which in turn uses the `ValidatorPluginManager`, which extends the `Zend\ServiceManager\AbstractPluginManager`. This class is not loaded because it is not part of the dependency.

To fix this issue  `zendframework/zend-servicemanager` needs to be a dependency, and not dev only.

froschdesign:
Please look at the suggest section in the Composer file:
https://github.com/zendframework/zend-inputfilter/blob/9569f7e18122d57b66c95ff2f02472937f26db46/composer.json#L26-L28

Zebercet:
I think this issue is not invalid. Because it's not a sugggest. You **have** to install servicemanager to use inputfilter

Xerkus:
Direct dependency was introduced in #67

settermjd:
@Xerkus, imo @Zebercet is correct. While it is introduced as a direct dependency in #67, zend-inputfilter/composer.json only suggests it. To me that's incorrect functionality. zend-servicemanager should be moved to be a core requirement.

voku:
I used the package yesterday for the first time and got this strange error ... so after a quick google-search I found this thread and now I know that I need to add "zend-servicemanager", but it isn't intuitive :/

PR #160, opened 2018-01-12
Add zend-servicemanager as a required dependency

settermjd:
Provide a narrative description of what you are trying to accomplish:

After being bitten by the bug in #119, where I couldn't use `zend-inputfilter` without `zend-servicemanager`, I'm proposing this PR, which makes `zend-servicemanager` a required dependency of `zend-inputfilter`. I can't claim to understand all the reasoning for not doing so to date (I'm happy to defer if necessary), but based on this experience it makes sense to make the change.

- [X] Are you fixing a bug?
  - [ ] Detail how the bug is invoked currently.
  - [ ] Detail the original, incorrect behavior.
  - [ ] Detail the new, expected behavior.
  
The above three questions are answered in https://github.com/zendframework/zend-inputfilter/issues/119#issue-172966930.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,15 +32,12 @@
         "php": "^5.6 || ^7.0",
         "zendframework/zend-filter": "^2.6",
         "zendframework/zend-stdlib": "^2.7 || ^3.0",
-        "zendframework/zend-validator": "^2.10.1"
+        "zendframework/zend-validator": "^2.10.1",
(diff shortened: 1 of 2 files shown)


settermjd:
Thanks for sharing the extra background information, @weierophinney. It helps me understand it in a greater context. And thanks for merging the PR too.

Issue #159, opened 2018-01-03
Null value for inputfilter causes exception


MadeRelevant:
Providing NULL as the value of nested input filter causes an exception to be thrown.
At the moment I only have my phone to post this issue so I will make a picture of the code to reproduce.
I am in the middle of moving and my ISP messed up :(

If you ommit the value its ok because an empty array will be used. I expected the same behaviour for NULL values.. Please tell me if I miss something here and how this should be fixed at my side in the case I did miss something.
![20180103_153355](https://user-images.githubusercontent.com/25172232/34524348-8e483ad8-f09b-11e7-90ff-2d75d4c86263.jpg)

MadeRelevant:
Passing nested => [] also works btw, which is clear from the code in the BaseInputFilter

MadeRelevant:
@weierophinney any clues?

froschdesign:
@MadeRelevant 
Please add the code of your test case here or create a PR. Copy & paste from a photo is a little bit complicated. 😉

weierophinney:
This is by design, currently. `InputFilter::setData()` expects either an array or a traversable set; any other data type results in an exception.

If you feel this should not be the case, please submit a pull request to allow passing null values.

Thanks!

MadeRelevant:
I will post code to reproduce tomorrow.
Of this is by design  i think it is strange  that a user is capabel of producing an exception.. Am i responsible to iterate over all data before sending it into the input filter?

weierophinney:
@MadeRelevant Let me flip the question around: why are you allowing a `null` value to be passed for that element?

And, related to that: what should the behavior of the input filter be when the data set is set to `null`, exactly?

MadeRelevant:
Thanks. It is not exactly that im allowing them to enter a null value but someone can. 

I think the behavior should be the same as if they didnt enter a value at all. Our API client doesnt allow it actually but the issue came forward out of our tests.

MadeRelevant:
@weierophinney sorry for the delay but here is the code of the screenshot:

```php
public function testBrokenInputFilter() {
        $filter1 = new InputFilter();
        $filter1->add([
            'type' => InputFilter::class,
            'nestedField1' => [
                'required' => false
            ]
        ], 'nested');
        $filter1->setData([]);
        self::assertNull($filter1->getValues()['nested']['nestedField1']); // <-- works

        // this is a value that a client may send when they try to break the API or
        // just don't know what to use as value for example. atm it produces an exception
[...]

PR #163, opened 2018-02-14
Hotfix/non iterable inputfilter value

MadeRelevant:
Fixes issue #159 

Currently, when you send a non-iterable value, non-null value to a nested input filter, it incorrectly raises an InvalidArgumentException. It should report as if all underlying fields are empty, and treat the value as an empty array.

### DIFF
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -178,15 +178,22 @@ class BaseInputFilter implements
      */
     public function setData($data)
     {
+        // A null value indicates an empty set
+        if (null === $data) {
+            $data = [];
+        }
+
         if ($data instanceof Traversable) {
             $data = ArrayUtils::iteratorToArray($data);
         }
 
-        // A null value indicates an empty set
-        if (null === $data || ! is_array($data)) {
-            $data = [];
+        if (! is_array($data)) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                '%s expects an array or Traversable argument; received %s',
+                __METHOD__,
+                (is_object($data) ? get_class($data) : gettype($data))
+            ));
         }
-
         $this->data = $data;
         $this->populate();
         return $this;
@@ -516,6 +523,14 @@ class BaseInputFilter implements
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
+                if ($value instanceof Traversable) {
+                    $value = ArrayUtils::iteratorToArray($value);
+                }
+
+                if (! is_array($value)) {
+                    $value = [];
+                }
+
                 $input->setData($value);
                 continue;
             }
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -523,11 +523,8 @@ class BaseInputFilter implements
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
-                if ($value instanceof Traversable) {
-                    $value = ArrayUtils::iteratorToArray($value);
-                }
-
-                if (! is_array($value)) {
+                // Fixes #159
+                if (! is_array($value) && !$value instanceof Traversable) {
                     $value = [];
                 }
 
--- a/src/BaseInputFilter.php
+++ b/src/BaseInputFilter.php
@@ -178,22 +178,15 @@ class BaseInputFilter implements
      */
     public function setData($data)
     {
-        // A null value indicates an empty set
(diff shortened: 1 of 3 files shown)


src/BaseInputFilter.php
```diff
@@ -523,6 +523,11 @@ protected function populate()
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
+                // Fixes #159
+                if (! is_array($value) && !$value instanceof Traversable) {
```
samsonasik: add space after not (`!`) operator

MadeRelevant: thanks for pointing that out, I saw the travis build fail indeed. Didn't find time to fix it yet.

MadeRelevant:
@samsonasik everything ready to merge?

samsonasik:
I’m not maintainer ^^

MadeRelevant:
> I’m not maintainer ^^

Lol, my bad

MadeRelevant:
@weierophinney ready to merge or do you need me something to change?

src/BaseInputFilter.php
```diff
@@ -523,6 +523,11 @@ protected function populate()
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
+                // Fixes #159
+                if (! is_array($value) && ! $value instanceof Traversable) {
```
svycka: I am not sure this is a good fix. In case of invalid data, validation result can be valid in some situations.

froschdesign: @svycka 
> …validation result can be valid in some situations.

In this case there is no validation result. Only an exception.

svycka: @froschdesign I understand that this fixes exception just not sure if this will not cause other problems. For example, if this nested input filter is optional and invalid data was given it will replace an invalid object with [] and nested input filter will not be able to validate or return an error in case of invalid data. 

But I don't know maybe that's not a problem since it was invalid anyway. If we ignore a problem that we can't return error messages for this then this is a good fix.

svycka:
@MadeRelevant your test does not cover your fix see: https://coveralls.io/builds/15520340/source?filename=src%2FBaseInputFilter.php#L528

src/BaseInputFilter.php
froschdesign: > if this nested input filter is optional

You can not set an input-filter as optional, therefore an exception is thrown.

svycka: okay, I am not sure about this so let's be it if this will be tested in all situations

froschdesign: Btw. for an optional input-filter you can use "[Optional Input Filters](https://docs.zendframework.com/zend-inputfilter/optional-input-filters/)".

svycka: I am more concerned not about optional input filter but about its optional inputs but I don't know if this is bad

froschdesign: The previous link is for an entire input-filter and not for single (optional) input!

MadeRelevant:
@svycka hmm, thats really weird, I ran the tests and it actually hits a breakpoint if I put it there.. Should be enough to prove it is touched and covered? When I run the tests with coverage it is for some strange reason not covered.. I have no clue why not while it reaches the breakpoit. I must be doing something wrong but not sure what, never had this before.

Any clues?

svycka:
don't know but if the test fails without this fix and does not with it then it is ok

MadeRelevant:
> don't know but if the test fails without this fix and does not with it then it is ok

Yes it fails without it. But still I think it is strange.. Bug in xdebug? PHPUnit reporter? Very weird..

svycka:
@MadeRelevant oh I see whats wrong you have to move test to BaseInputFilterTest or add annotation like this:
````php
    /**
     * @covers \Zend\InputFilter\BaseInputFilter
     */
    public function testNestedInputFilterShouldAllowNonArrayValueForData()
    {
// ...
````

File: src/BaseInputFilter.php [hunk 3]
```diff
@@ -523,6 +523,11 @@ protected function populate()
             $value = $this->data[$name];
 
             if ($input instanceof InputFilterInterface) {
+                // Fixes #159
```
svycka: I guess comment is not required as it is covered with test

MadeRelevant: Nice, coverage is fixed. I left the comment where it was. If it needs to be removed tell me.

PR #165, opened 2018-04-16
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] normalizes `composer.json`

💁‍♂️ For reference, see https://github.com/localheinz/composer-normalize.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,12 +1,23 @@
 {
     "name": "zendframework/zend-inputfilter",
     "description": "Normalize and validate input sets from the web, APIs, the CLI, and more, including files",
-    "license": "BSD-3-Clause",
     "keywords": [
         "zf",
         "zendframework",
         "inputfilter"
     ],
+    "license": "BSD-3-Clause",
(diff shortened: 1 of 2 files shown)


froschdesign:
@localheinz 
I think, you should create a suggestion at [zendframework/maintainers](https://github.com/zendframework/maintainers), because your plugin would be a help for all components.

localheinz:
@froschdesign 

> I think, you should create a suggestion at [`zendframework/maintainers`](https://github.com/zendframework/maintainers), because your plugin would be a help for all components.

Thank you! What do you suggest, opening an issue, or a pull request?

weierophinney:
We actually _already_ have a tool in the maintainers repo for normalization. However, instead of putting the keys in alphabetical order, we have been grouping them based on functionality (e.g., `require`, `require-dev`, `conflict`, and `suggest` are grouped, in that order). IIRC, we already did that in this repo, so I'm going to close this issue; we can revisit if we decide to order them differently later.

PR #167, opened 2018-05-17
Required does not work, when array is empty

eweso:
The first problem is, that hasValue always returns true, when a value is set, even, if the value is an empty array. So the first validation for the required filter will be skipped. The second problem is, that the isRequired is called within the foreach loop. But if the value is an empty array, the loop will never start, so isRequired will never triggered and the method will return the defaultValue true.

### DIFF
--- a/src/ArrayInput.php
+++ b/src/ArrayInput.php
@@ -83,6 +83,14 @@ class ArrayInput extends Input
         $validator = $this->getValidatorChain();
         $values    = $this->getValue();
         $result    = true;
+        
+        if ($required && empty($values)) {
+            if ($this->errorMessage === null) {
+                $this->errorMessage = $this->prepareRequiredValidationFailureMessage();
+            }
+            return false;
+        }
+        
         foreach ($values as $value) {
             $empty = ($value === null || $value === '' || $value === []);
             if ($empty && ! $this->isRequired() && ! $this->continueIfEmpty()) {
--- a/src/ArrayInput.php
+++ b/src/ArrayInput.php
@@ -83,14 +83,14 @@ class ArrayInput extends Input
         $validator = $this->getValidatorChain();
         $values    = $this->getValue();
         $result    = true;
-        
+
         if ($required && empty($values)) {
             if ($this->errorMessage === null) {
                 $this->errorMessage = $this->prepareRequiredValidationFailureMessage();
             }
             return false;
         }
-        
+
         foreach ($values as $value) {
             $empty = ($value === null || $value === '' || $value === []);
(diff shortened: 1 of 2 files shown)


eweso:
@webimpress Thank you for your feedback. I was on holiday until now. The point is, that I am really new to github and don't know, where the test files are and where I have to create one, to pass the checks. If someone could help me a little, we could fix that bug quickly.

weierophinney:
@eweso The repository contains information on contributing, including how to run tests: https://github.com/zendframework/zend-inputfilter/blob/master/docs/CONTRIBUTING.md
(BTW, tests are in the `test/` subdirectory.)

If you need help writing tests, jump on the [Slack](https://zendframework-slack.herokuapp.com) and join the #contributors channel and start asking.

eweso:
Okay, I still do not get, why the first test failed. There was already a test class in this repository. The basic behaviour hasn't changed. When I open the Details of coverage/coveralls it does not give me any information about missing test classes. 

It just informs me, that some lines of codes have changed in the base repository and that Module.php ist missing. So am I right, that the coverage/coveralls does not fail because of the test, it fails because some lines of code were changed in the base branch, to which I wanted to pull my request?

If so, I think, I just have to update those files in my patch and the test will be successful?

rkeet:
Hi, sorry to pitch in here. I've got a similar but reverse check in a repo of mine, however I've done it recursively. Would that be worth checking out? 

Function used [here](https://github.com/rkeet/zf-doctrine-form/blob/8974de7ca21eae290580c15ec445128fe1cfad88/src/InputFilter/AbstractInputFilter.php#L58)
Modified isValid [here](https://github.com/rkeet/zf-doctrine-form/blob/8974de7ca21eae290580c15ec445128fe1cfad88/src/InputFilter/AbstractInputFilter.php#L88)

Checks if `input` is **not** required _and_ if that `input` is recursively empty, then return true. Includes checking nested inputs and accounts for nested Fieldsets and Collections.

Could be this comes to the same result, but in the committed code I do not see if it takes into account nested Fieldsets and Collections, which is why I thought to mention it so that it at least gets looked at. 

Currently do not have the time to test this PR.

eweso:
@rkeet

This is a different case. I fixed a bug in the ArrayInput. You are talking about the InpuFilter.

When you have a Multi-Select for example and you use an ArrayInput with required in your InputFilter to validate the Multi-Select and there is no value selected, the ArrayInput returns at the moment true, because it does not handle the empty array as an empty value.

rkeet:
I was indeed, this is a different case, my apologies. Removed comment.

weierophinney:
Thanks, @eweso!

Issue #168, opened 2018-06-19
collectionRawValues not populated in setData of CollectionInputFilter


rkeet:
Narrative:
I use the following `isValid` function in my InputFilter class:

    public function isValid($context = null)
    {
        if ( ! $this->isRequired() && $this->arrayFilterIsEmpty($this->getRawValues())) {

            return true;
        }

        return parent::isValid();
    }

The `function arrayFilterIsEmpty()` checks whether or not the current `$input` contains any data and is required. The `parent::isValid()` calls the normal isValid function.

Any input (normal Input or Collection) has the function `getRawValues` to get the raw input data. I use it to check if any data is present combined with a required property to see if a both not-required and empty fieldset/collection can be skipped during validation and just be marked as valid. 

During `isValid` in `CollectionInputFilter.php` ([line 195](https://github.com/zendframework/zend-inputfilter/blob/4901fc67554efbb7c292ae499129792977df1220/src/CollectionInputFilter.php#L195)) the function `setData` is called ([here](https://github.com/zendframework/zend-inputfilter/blob/4901fc67554efbb7c292ae499129792977df1220/src/CollectionInputFilter.php#L140)). 

However, `setData` breaks of setting data if the data is an array or Traversable ([here](https://github.com/zendframework/zend-inputfilter/blob/4901fc67554efbb7c292ae499129792977df1220/src/CollectionInputFilter.php#L151)). This causes the `collectionRawValues` property to never be set.

As such, I suggest updating the function to set this value after the initial check, where it's still handling the collection as a whole instead of per item (before the `foreach()`)

    public function setData($data)
    {
        if ( ! (is_array($data) || $data instanceof Traversable)) {
            throw new InvalidArgumentException(
                sprintf(
                    '%s expects an array or Traversable collection; invalid collection of type %s provided',
                    __METHOD__,
                    is_object($data) ? get_class($data) : gettype($data)
                )
            );
        }

        $this->collectionRawValues = $data; // added line 

        foreach ($data as $item) {
            if (is_array($item) || $item instanceof Traversable) {
                continue;
            }

            throw new InvalidArgumentException(
                sprintf(
                    '%s expects each item in a collection to be an array or Traversable; '
                    . 'invalid item in collection of type %s detected',
                    __METHOD__,
                    is_object($item) ? get_class($item) : gettype($item)
                )
            );
        }

        $this->data = $data;
        return $this;
    }

### Code to reproduce the issue

[...]

PR #169, opened 2018-06-19
Added setting of collectionRawValues property on CollectionInputFilter during setData

rkeet:
Provide a narrative description of what you are trying to accomplish:

- [x] Are you fixing a bug?

Fix for bug described here: https://github.com/zendframework/zend-inputfilter/issues/168

### DIFF
--- a/src/CollectionInputFilter.php
+++ b/src/CollectionInputFilter.php
@@ -147,6 +147,8 @@ class CollectionInputFilter extends InputFilter
             ));
         }
 
+        $this->collectionRawValues = $data;
+
         foreach ($data as $item) {
             if (is_array($item) || $item instanceof Traversable) {
                 continue;


rkeet:
CS-Check passed
- It moaned about line endings (I'm on Windows) but didn't apply to commited file (commiting Unix style)

froschdesign:
@rkeet 
Please provide also an unit test. Thanks!

rkeet:
@froschdesign I gave it a shot. Does/would it suffice?

rkeet:
@froschdesign Could you have a look and let me know what you think?
@weierophinney pinging by suggestion of @svycka to see if you could possibly also have a look and give an opinion?

froschdesign:
@rkeet 
I think, we should look at the behaviour of the `BaseInputFilter` class and compare:

```php
public function testRawValuesMethod()
{
    $data = [
        'field' => 'foo bar',
        'raw'   => 'baz',
    ];

    // Input-filter
    $inputFilter = new BaseInputFilter();
    $input = new Input('field');
    $input->getFilterChain()->attachByName(SeparatorToDash::class);
    $inputFilter->add($input);
    $inputFilter->setData($data);

    // Tests
    $this->assertEquals(['field' => 'foo bar'], $inputFilter->getRawValues());

    $this->assertTrue($inputFilter->isValid());

    $this->assertArrayHasKey('field', $inputFilter->getValues());
    $this->assertArrayNotHasKey('raw', $inputFilter->getValues());
    $this->assertArrayHasKey('field', $inputFilter->getRawValues());
    $this->assertArrayNotHasKey('raw', $inputFilter->getRawValues());

    $this->assertEquals(['field' => 'foo-bar'], $inputFilter->getValues());
    $this->assertEquals(['field' => 'foo bar'], $inputFilter->getRawValues());
}
```
The raw value(s) means the _unfiltered_ value(s) – not the entire input of `setData`.

### References

#### Code

https://github.com/zendframework/zend-inputfilter/blob/4901fc67554efbb7c292ae499129792977df1220/src/Input.php#L286-L289

https://github.com/zendframework/zend-inputfilter/blob/4901fc67554efbb7c292ae499129792977df1220/src/Input.php#L313-L317

https://github.com/zendframework/zend-inputfilter/blob/4901fc67554efbb7c292ae499129792977df1220/src/BaseInputFilter.php#L433-L441

#### Documentation

> The `getValue()` method returns the filtered value of the 'foo' input, while `getRawValue()` returns the original value of the input.

https://docs.zendframework.com/zend-inputfilter/intro/#introduction

---

So this is wrong: `'raw_only' => 'not in filters - should be in raw data',`

(I hope I understood your goal correctly.)

svycka:
@froschdesign I guess you did not understand he changed how raw values will be retrieved. Before this change raw values were retrieved from each inputFilter with their getRawValues() logic. and now raw values are set in setData() method and it does not do anything with it just sets what it gets and later returns.

example what could be different(pseudo code)
```php
// inputfilter config:
CollectionInputFilter
    inputFilter
       - name

// example input data
[
  'collection' => [
    ['name' => 'your name', 'invalid_field' => 'omg'],
  ]
]
// getRawValues() result before this PR
[
    ['name' => 'your name'],
]
// getRawValues() result after this PR
[
    ['name' => 'your name', 'invalid_field' => 'omg'],
]
```
but after this PR getRawValues should work a lot faster :D as it only does one set and one get comparing to `collection_elements_count*collection_inputfiler_fields_count` calls and it't strange why it was implemented like that and because of this I think there was a reason for this.

froschdesign:
@svycka 
I think, you missed the key question: Why should the invalid field be included?

> ```php
> // getRawValues() result after this PR
> [
>     ['name' => 'your name', 'invalid_field' => 'omg'],
> ]
> ```

svycka:
@froschdesign yes maybe you know the answer?
also, I am not sure there are no more differences because raw values now get from a lot of inputs and foreach... where can have different getRawValues implementation

froschdesign:
@svycka 
> yes maybe you know the answer?

See my comment above:

> The raw value(s) means the unfiltered value(s) of defined input(s) – not the entire input of `setData`.

I also see the problem of a BC break and the changes in the current behaviour, because the return values are different and the raw values are set before validation. If we include the entire input of `setData` then we hijack the `getRawValues` methods.

I think, we should wait for an answer from @rkeet.

froschdesign:
@rkeet 
Thanks for your answer! 👍 

But that still does not make it right:

```php
// getRawValues() result after this PR
[
    ['name' => 'your name', 'invalid_field' => 'omg'],
]
```

> If anything, this fix caused multiple issues to be resolved, such as re-populating forms for continued editing, validation of deep nesting (BaseFieldset > Fieldset > Collection > Fieldset > Collection > Fieldset), and a host of other small issues which cause known headaches.

Please don't get me wrong, it is very good to fix these problem, but changing the return value of `getRawValues` method is wrong here.

rkeet:
> [...] changing the return value of getRawValues method is wrong here.

After further considering this, yes, I agree. Also, I'm having trouble thinking of a proper fix, because setting this data here, does take care of the issues I've mentioned, at least when using such deep nesting as exampled. 

I'm open to suggestions here. Apart from a version bump with this fix, I'm flush out of ideas on for this issue at the moment.

weierophinney:
I've had a chance to read through this and the original issue (#168), and have the following observations:

- As @froschdesign notes, `getRawValues()` and `getRawValue()` are expected to return values only for _known_ inputs. For collections, this means descending into each group in the collection and fetching those. I'm not clear after the various discussions if this is happening or not; if it isn't, we need to fix it.

- We already have a method for retrieving _unknown_ inputs: `getUnknown()`. I'm not sure if this is useful for collections, however.

- Following on the above two points, I can see a clear case for being able to access the _full set of data passed to a collection_. As @rkeet noted in the original issue, this could be used to vary validation based on what fields are present.

My take is that we should add:

```php
interface UnfilteredDataInterface
{
    /**
     * @return array
     */
    public function getUnfilteredData();
```

We would implement this interface in input filters and collections, and they would essentially just return the value of the `$data` property.

If this seems reasonable, let's go ahead and start working on that aspect. In the meantime, @rkeet , can you indicate whether or not `getRawValues()` is broken for collections (given the guidelines I outlined above), and, if so, write up a unit test demonstrating that? I can potentially work on solutions to that and/or the `UnfilteredDataInterface` additions early next week.

svycka:
@weierophinney is this `getUnfilteredData();` will be used as `$context` in validators?

froschdesign:
@svycka 
I think this would change the current behaviour and ends in a BC break.

https://github.com/zendframework/zend-inputfilter/blob/799ad48ed1666d3c62126fec73dd20453b3a9e4d/src/BaseInputFilter.php#L232

svycka:
then why would we need `getUnfilteredData()` as it will be impossible to use in validation

froschdesign:
@svycka 
We will wait for the answer / solution from @rkeet. I think he already has an idea for it.

rkeet:
@froschdesign @svycka @weierophinney 

I haven't forgotten this, just didn't have time this weekend. I hope to get to it this week either during work, an evening or maybe during the coming weekend. However, very busy at the moment. 

What I'll try to do the coming days:

* Setup up a skeleton application demonstrating issue with plain modules, comments, etc. 
* Summarize this thread & [#168](https://github.com/zendframework/zend-inputfilter/issues/168) into a coherent whole 

---

Quick re-read though:

* `getRawValue(s)()` should _not_ be altered from existing functionality to prevent bc-break
* Agreed with that `getRawValue(s)()` returns only _known_ input values, `UnfilteredDataInterface` should provide solution next to these functions
* `UnfilteredDataInterface`  adds functionality instead of modifying existing ones

---

I'll get into that more by providing the code I mentioned with unit testing - I hope this can wait a few more days (possibly weeks). Like I said, currently very busy, also next to my job. Though holidays soon, so might get something done then...

Issue #94, opened 2016-02-12
Missing required failure message for CollectionInputFilter


Maks3w:
CollectionInputFilter fails without error message when is required but data is empty or not set.

n3vrax:
Is this fixed? Using the latest version, and it seems indeed that the collection input filter does not have some sort of default message when required is true and an empty array is given. Need to extend the input filter for now.

developer-devPHP:
Issue is still exists. @weierophinney  can you confirm this?

developer-devPHP:
Any updates on this? Problem still exists.

weierophinney:
Please provide us with a minimal coffee sample that demonstrates the issue; right now, we'd be doing guess work.

developer-devPHP:
# Example Input Filter
```php
$factory     = new \Zend\InputFilter\Factory();
$inputFilter = $factory->createInputFilter(
[
    'collection-element' => [
		'type'         => \Zend\InputFilter\CollectionInputFilter::class,
		'required'     => true,
		'input_filter' => [
			'type'  => [
				'required'     => true,
				'validators' => [
					[
						'name'    => \Zend\Validator\Between::class,
						'options' => [
							'min' => 50,
							'max' => 100,
						],
					],
				],
			],
			'price' => [
				'required'     => true,
				'validators' => [
					[
						'name'    => \Zend\Validator\Between::class,
						'options' => [
							'min' => 50,
							'max' => 100,
						],
					],
				],
			],
		],
	],
]
);
```
## First Try, when you have a problem with empty error message
```php
$inputFilter->setData([]);
var_dump($inputFilter->isValid());
var_dump($inputFilter->getMessages());
```
## Second Try with Valid error messages if provided some data
```php
$inputFilter->setData([
	'collection-element' => [
		0 => [

		],
	],
]);
var_dump($inputFilter->isValid());
var_dump($inputFilter->getMessages());
```

froschdesign:
Output is:

### First try
```
boolean false

array (size=1)
  'collection-element' => 
    array (size=0)
      empty
```

### Second try
```
false

array (size=1)
  'collection-element' => 
    array (size=1)
      0 => 
        array (size=2)
          'type' => 
            array (size=1)
              'isEmpty' => string 'Value is required and can't be empty' (length=36)
          'price' => 
            array (size=1)
              'isEmpty' => string 'Value is required and can't be empty' (length=36)
```

froschdesign:
No error message is generated.

[...]

PR #170, opened 2018-07-18
fixxed #94 'Missing required failure message for CollectionInputFilter' issue

ruzann:
Added default message to display when a collection is required.
Added mechanism for setting a custom "required" message.
Updated  factorie to allow setting that message.

### DIFF
--- a/src/CollectionInputFilter.php
+++ b/src/CollectionInputFilter.php
@@ -10,6 +10,7 @@
 namespace Zend\InputFilter;
 
 use Traversable;
+use Zend\Validator\NotEmpty;
 
 class CollectionInputFilter extends InputFilter
 {
@@ -42,6 +43,11 @@ class CollectionInputFilter extends InputFilter
      * @var BaseInputFilter
      */
     protected $inputFilter;
+
+    /**
+     * @var NotEmpty
+     */
+    protected $notEmptyValidator;
 
     /**
      * Set the input filter to use when looping the data
@@ -165,6 +171,30 @@ class CollectionInputFilter extends InputFilter
     }
 
     /**
+     * @return NotEmpty
+     */
+    public function getNotEmptyValidator()
+    {
+        if ($this->notEmptyValidator === null) {
+            $this->notEmptyValidator = new NotEmpty();
+        }
+
+        return $this->notEmptyValidator;
+    }
+
+    /**
+     * @param NotEmpty $notEmptyValidator
+     *
+     * @return $this
+     */
+    public function setNotEmptyValidator(NotEmpty $notEmptyValidator)
+    {
+        $this->notEmptyValidator = $notEmptyValidator;
+
+        return $this;
+    }
+
+    /**
      * {@inheritdoc}
      * @param mixed $context Ignored, but present to retain signature compatibility.
      */
@@ -176,6 +206,7 @@ class CollectionInputFilter extends InputFilter
 
         if ($this->getCount() < 1) {
             if ($this->isRequired) {
+                $this->collectionMessages[] = $this->prepareRequiredValidationFailureMessage();
                 $valid = false;
             }
         }
@@ -294,5 +325,24 @@ class CollectionInputFilter extends InputFilter
         }
 
         return $unknownInputs;
+    }
+
+    /**
+     * @return array
+     */
+    protected function prepareRequiredValidationFailureMessage()
+    {
+        $notEmptyValidator = $this->getNotEmptyValidator();
+        $templates  = $notEmptyValidator->getOption('messageTemplates');
+        $message    = $templates[NotEmpty::IS_EMPTY];
+        $translator = $notEmptyValidator->getTranslator();
+
+        if ($translator) {
+            $message = $translator->translate($message, $notEmptyValidator->getTranslatorTextDomain());
+        }
+
+        return [
+            NotEmpty::IS_EMPTY => $message,
+        ];
     }
 }
--- a/src/CollectionInputFilter.php
+++ b/src/CollectionInputFilter.php
@@ -171,6 +171,11 @@ class CollectionInputFilter extends InputFilter
     }
 
     /**
+     * Retrieve the NotEmpty validator to use for failed "required" validations.
+     *
+     * This validator will be used to produce a validation failure message in
+     * cases where the collection is empty but required.
+     *
      * @return NotEmpty
      */
     public function getNotEmptyValidator()
@@ -183,8 +188,12 @@ class CollectionInputFilter extends InputFilter
     }
 
     /**
-     * @param NotEmpty $notEmptyValidator
+     * Set the NotEmpty validator to use for failed "required" validations.
      *
+     * This validator will be used to produce a validation failure message in
+     * cases where the collection is empty but required.
+     *
+     * @param NotEmpty $notEmptyValidator
      * @return $this
      */
     public function setNotEmptyValidator(NotEmpty $notEmptyValidator)
@@ -204,11 +213,9 @@ class CollectionInputFilter extends InputFilter
         $inputFilter = $this->getInputFilter();
         $valid = true;
 
-        if ($this->getCount() < 1) {
-            if ($this->isRequired) {
-                $this->collectionMessages[] = $this->prepareRequiredValidationFailureMessage();
-                $valid = false;
-            }
+        if ($this->getCount() < 1 && $this->isRequired) {
+            $this->collectionMessages[] = $this->prepareRequiredValidationFailureMessage();
+            $valid = false;
         }
 
         if (count($this->data) < $this->getCount()) {
@@ -328,21 +335,19 @@ class CollectionInputFilter extends InputFilter
     }
 
     /**
-     * @return array
+     * @return array<string, string>
      */
     protected function prepareRequiredValidationFailureMessage()
     {
         $notEmptyValidator = $this->getNotEmptyValidator();
-        $templates  = $notEmptyValidator->getOption('messageTemplates');
-        $message    = $templates[NotEmpty::IS_EMPTY];
-        $translator = $notEmptyValidator->getTranslator();
-
-        if ($translator) {
-            $message = $translator->translate($message, $notEmptyValidator->getTranslatorTextDomain());
-        }
+        $templates         = $notEmptyValidator->getOption('messageTemplates');
+        $message           = $templates[NotEmpty::IS_EMPTY];
+        $translator        = $notEmptyValidator->getTranslator();
 
         return [
-            NotEmpty::IS_EMPTY => $message,
+            NotEmpty::IS_EMPTY => $translator
+                ? $translator->translate($message, $notEmptyValidator->getTranslatorTextDomain())
+                : $message,
         ];
     }
 }
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -319,6 +319,9 @@ class Factory
             if (isset($inputFilterSpecification['required'])) {
                 $inputFilter->setIsRequired($inputFilterSpecification['required']);
             }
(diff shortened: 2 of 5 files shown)


src/CollectionInputFilter.php
```diff
@@ -176,6 +206,14 @@ public function isValid($context = null)
 
         if ($this->getCount() < 1) {
             if ($this->isRequired) {
+
+                if($this->messageRequired == null)
+                {
```
malukenho: Code-standards isn't right

src/CollectionInputFilter.php
```diff
@@ -176,6 +206,14 @@ public function isValid($context = null)
 
         if ($this->getCount() < 1) {
             if ($this->isRequired) {
+
+                if($this->messageRequired == null)
```
malukenho: May we use strict comparison here?

src/CollectionInputFilter.php
```diff
@@ -43,6 +45,11 @@ class CollectionInputFilter extends InputFilter
      */
     protected $inputFilter;
 
+    /**
+     * @var $messageRequired
```
malukenho: It should be the type of the property not their label

src/CollectionInputFilter.php
```diff
@@ -176,6 +205,10 @@ public function isValid($context = null)
 
         if ($this->getCount() < 1) {
             if ($this->isRequired) {
+                if ($this->messageRequired === null) {
+                    $this->messageRequired = self::NOT_EMPTY;
+                }
+                $this->collectionMessages[] = $this->messageRequired;
```
malukenho: Does it need to change the state of `$this->messageRequired`? 
We could do something like this:

```php
$this->collectionMessages[] = $this->messageRequired ?: self::NOT_EMPTY
```

malukenho: In that case you should use `static::CONSTANT` instead.

src/CollectionInputFilter.php
```diff
@@ -164,6 +171,28 @@ public function setData($data)
         return $this;
     }
 
+    /**
+     * Set "required"  message if collection can not be empty
+     *
+     * @param string $messageRequired
+     * @return CollectionInputFilter
+     */
+    public function setMessageRequired($messageRequired)
```
froschdesign: Unit tests for this method is still missing. If some tests have been added, you will notice that a validation for a string is missing and that also `null` is allowed.

src/CollectionInputFilter.php
```diff
@@ -164,6 +171,28 @@ public function setData($data)
...
+
+        return $this;
+    }
+
+    /**
+     *  Get "required"  message if collection can not be empty
+     *
+     * @return string
+     */
+    public function getMessageRequired()
```
froschdesign: The same here: unit test is missing.

src/CollectionInputFilter.php
```diff
@@ -43,6 +45,11 @@ class CollectionInputFilter extends InputFilter
      */
     protected $inputFilter;
 
+    /**
+     * @var string
```
froschdesign: Incomplete – see my previous comment.

test/CollectionInputFilterTest.php
```diff
@@ -735,4 +735,32 @@ public function testDuplicatedErrorMessages()
...
+        $value = null;
+        $this->inputFilter->setMessageRequired($value);
+        $this->assertEquals($value, $this->inputFilter->getMessageRequired());
+    }
+
+    public function messageRequiredProvider()
+    {
+        return [
+            ['foo'],
+            [null],
```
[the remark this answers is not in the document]
froschdesign: `array` and `object` are missing. You also have to test all values that are not expected!

ruzann:
@froschdesign Is there any updates relation to pull request?

froschdesign:
@ruzann 
I still see a problem here, because your changes introduce a hard coded error message to an input-filter, that means we have message constants in input-filters _and_ validators.
This ends in another problem, because the error message can not be translated.

src/CollectionInputFilter.php
```diff
@@ -295,4 +333,38 @@ public function getUnknown()
...
+     * @param array|null $options
+     *
+     * @return array
+     */
+    protected function prepareRequiredValidationFailureMessage(array $options = null)
+    {
+        $notEmpty = new NotEmpty($options);
+        $templates  = $notEmpty->getOption('messageTemplates');
+        $message    = $templates[NotEmpty::IS_EMPTY];
+        $translator = new Translator();
```
froschdesign: Compare with with the `Input` class:

https://github.com/zendframework/zend-inputfilter/blob/4901fc67554efbb7c292ae499129792977df1220/src/Input.php#L511

composer.json
```diff
@@ -34,7 +34,7 @@
         "zendframework/zend-stdlib": "^2.7 || ^3.0",
         "zendframework/zend-validator": "^2.10.1",
         "zendframework/zend-servicemanager": "^2.7.10 || ^3.3.1",
-        "zendframework/zend-i18n": "^2.9.0",
+        "zendframework/zend-i18n": "^2.9.0"
```
froschdesign: Please revert this and see my previous comment!

froschdesign:
@ruzann 
Do not create a new translator object, this will introduce new problems and you must set the the translation files for the input-filter - this is wrong!
Please revert this and use the translator from the validator.

src/CollectionInputFilter.php
```diff
@@ -10,9 +10,12 @@
 namespace Zend\InputFilter;
 
 use Traversable;
+use Zend\Validator\NotEmpty;
```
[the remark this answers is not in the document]
ruzann: class is already removed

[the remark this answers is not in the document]
developer-devPHP: Thank you

ruzann:
@froschdesign all your notes fixed, please check again, when you have a time.

newdevonair:
@froschdesign is there any news from this pull request?
I need this functionality in my project too.

Thanks.

froschdesign:
@ruzann 
Some pseudo code:

```php
function getNotEmptyValidator()
{
    if ($this->notEmptyValidator === null) {
        $this->notEmptyValidator = new NotEmpty();
    }

    return $this->notEmptyValidator;
}

function setNotEmptyValidator($notEmptyValidator);

function prepareRequiredValidationFailureMessage();
```

The default error message already exists and you can replace the validator with your own. `messageRequired` and the related methods are not needed.

_(Yes, we do not use the validator for validation, but we allow a translation of the error message with the standard way.)_

/cc @malukenho @rkeet 

@newdevonair 
You can also add a review or a comment to the current suggestions. 😉 

---

Sorry for the late response and the short review. My focus is currently on the documentation, website and zend-navigation.

rkeet:
I would suggest same as @froschdesign review: own getter/setter for storing NotEmpty instance.

Otherwise seems alright to me, apart from tests not passing ;)

ruzann:
@froschdesign I fixed your notes, please check again.

developer-devPHP:
Hi guys, do you have any updates related this issue? Is it ready to merge or have any other improvement?
If you need some help I am ready to help.

froschdesign:
@weierophinney 
> Alternately, we should allow specifying a validator and/or validator chain at the top level for checking the status of the collection, prior to attempting to validate the members of the collection.

This one! (See my example with pseudo code above.)
Why? In all inputs we uses the messages from the validator. Setting a message to an input would introduce a new behaviour for this component. Setting a validator or a validator chain is not new.
The validator has already the option to translate a message.

ruzann:
@weierophinney, @froschdesign  is there any news? 
should I change something?

weierophinney:
@ruzann Please see the feedback above - @froschdesign and I feel that the approach outlined here is convoluted, and inconsistent with the architecture. Adding a validator chain specific to the collection input filter as an entity would make more sense. If you can refactor to that approach, we can review again.

src/CollectionInputFilter.php
```diff
@@ -164,6 +170,42 @@ public function setData($data)
...
+
+        return $this->notEmptyValidator;
+    }
+
+    /**
+     * @param string $notEmptyValidatorMessage
+     *
+     * @return $this
+     */
+    public function setNotEmptyValidator($notEmptyValidatorMessage)
```
froschdesign: This method is for setting the `NotEmpty` validator, not for the message. The message can be set on the validator itself:

```php
(new CollectionInputFilter)->getNotEmptyValidator()->setMessage('Please fill out this field.')
```

src/Factory.php
```diff
@@ -319,6 +319,9 @@ public function createInputFilter($inputFilterSpecification)
             if (isset($inputFilterSpecification['required'])) {
                 $inputFilter->setIsRequired($inputFilterSpecification['required']);
             }
+            if (isset($inputFilterSpecification['messageRequired'])) {
```
froschdesign: Please use `required_message` for consistency.

src/CollectionInputFilter.php
```diff
@@ -182,30 +182,6 @@ public function getNotEmptyValidator()
         return $this->notEmptyValidator;
     }
 
-    /**
-     * @param string $notEmptyValidatorMessage
-     *
-     * @return $this
-     */
-    public function setNotEmptyValidator($notEmptyValidatorMessage)
```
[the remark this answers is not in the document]
ruzann: done

ruzann:
@froschdesign is there any news?

froschdesign:
@ruzann 
Your improvement will be included in the next minor release (2.9). At the moment I am working at documentation of zend-hydrator, therefore I can not give a time frame. _I'm sorry._

In the meantime, just use your own fork.

test/CollectionInputFilterTest.php
```diff
@@ -735,4 +735,39 @@ public function testDuplicatedErrorMessages()
...
+            $this->inputFilter->getNotEmptyValidator()->getMessageTemplates()[NotEmpty::IS_EMPTY]
+        );
+
+        $value = 'foo';
+        $this->inputFilter->getNotEmptyValidator()->setMessage($value);
+        $this->assertEquals(
+            $value,
+            $this->inputFilter->getNotEmptyValidator()->getMessageTemplates()[NotEmpty::IS_EMPTY]
+        );
+    }
```
weierophinney: The above two tests are essentially identical. The behavior we need to test is:

- `getNotEmptyValidator()` lazy-loads an instance if none was previously set.
- `setNotEmptyValidator()` allows injecting a custom instance.
- When using `Factory`, we can provide a custom message via the `required_message` configuration setting.

I'll make those changes and push them to your branch shortly.

src/CollectionInputFilter.php
```diff
@@ -295,4 +326,23 @@ public function getUnknown()
...
+        $message    = $templates[NotEmpty::IS_EMPTY];
+        $translator = $notEmptyValidator->getTranslator();
+
+        if ($translator) {
+            $message = $translator->translate($message, $notEmptyValidator->getTranslatorTextDomain());
+        }
+
+        return [
+            NotEmpty::IS_EMPTY => $message,
+        ];
```
weierophinney: Lines 340-346 could be simplified by using a ternary to provide the message to the IS_EMPTY key.

I'll push that change to your branch shortly.

src/CollectionInputFilter.php
```diff
@@ -176,6 +206,7 @@ public function isValid($context = null)
 
         if ($this->getCount() < 1) {
             if ($this->isRequired) {
```
weierophinney: Since there's only one other condition we check when the count is less than one, we can combine these conditionals. I'll do that and push to your branch shortly.

weierophinney:
Thanks, @ruzann!

PR #172, opened 2018-08-16
Add file input compatibility with PSR-7

### LINKED ISSUE #145: FileInput not compatible with PSR (and Dictatoros) UploadedFile
Hello.

Zend\InputFilter\FileInput does not accept an implementation of Psr\Http\Message\UploadedFileInterface **if the input is required**, for exemple, i have this form with InputFilterProviderInterface :

```php
class MyForm extends Form implements InputFilterProviderInterface
{

    public function __construct()
    {
        parent::__construct();

        $this->add([
            'name' => 'myPicture',
            'type'  => 'File',
        ]);

        $this->add([
            'name' => 'submit',
            'type'  => 'Submit',
            'attributes' => [
                'value' => 'Submit',
                'class' => 'btn-primary',
            ],
        ]);
    }


    /**
      * @return array
      */
    public function getInputFilterSpecification() : array
    {
        return [
            'myPicture' => [
                'required' => true, // <========
            ],
        ];
    }
}
```

And in an Expressive middleware :

```php
$form = new MyForm;

if ($request->getMethod() === 'POST') {
    $form->setData(array_merge_recursive(
        // @var Psr\Http\Message\ServerRequestInterface $request
        $request->getParsedBody(),
        $request->getUploadedFiles()
    ));
    var_dump($form->isValid()); // Display always false...
}

```

alextech:
- [x] Are you creating a new feature?
  - [x] Why is the new feature needed? What purpose does it serve?
Fixes #145 - PSR-7 compatibility for validating and filtering forms with inputs.
  - [x] How will users use the new feature?
Validating forms submitting through PSR payload.
  - [x] Base your feature on the `develop` branch, and submit against that branch.
  - [x] Add only one feature per pull request; split multiple features over multiple pull requests
  - [x] Add tests for the new feature.
  - [x] Add documentation for the new feature.
Drop in replacement functions same way as in current docs.
  - [x] Add a `CHANGELOG.md` entry for the new feature.

**DO NOT MERGE COMPOSER.JSON AS IS**
This repository has tightly coupled dependencies with `zend-validator` repository, so I had to add my fork as a repository version with prerequisite PR. Merge it *first*.
Similar for `zend-filter` though that one is technically optional for `RenameUpload` filter, though useless without it and referenced in the docs.

**Prerequisite PRs**:
- Validation: https://github.com/zendframework/zend-validator/pull/237
- Filter: https://github.com/zendframework/zend-filter/pull/70

I could not separate PSR vs HTTP $_FilES into completely separate subclasses because current implementation is stateful, so cannot simply pass calls from current `FileInput` to appropriate implementation - private protected properties that hold values and options get lost. `FileInput` is hardcoded in other repositories, eg `Zend\Form` as a input specification so has to remain main entry point for forms to be used in both MVC and Expressive. Consequently, did decorator-ish style approach.


alextech:
Looks like Travis does not like to pull from other repositories not having its own tokens.
Can rerun the job after the 2 dependencies get accepted.

samsonasik:
@alextech you can insert `"no-api": true` config inside each "repositories" config record in composer.json as follow:

```php
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/alextech/zend-validator",
            "no-api": true
        },
        {
            "type": "vcs",
            "url": "https://github.com/alextech/zend-filter",
            "no-api": true
        }
    ],
```

src/FileInput.php
```diff
@@ -138,58 +144,21 @@ public function isValid($context = null)
...
+        return $this->implementation->isValid($context);
     }
 
     /**
-     * @return void
+     * @param  InputInterface $input
+     *
+     * @return FileInput
      */
-    protected function injectUploadValidator()
```
svycka: removing private methods isn't  BC break?

webimpress: This method is `protected` ;-) Only changes to __public__ interface are considered as BC Break.

alextech:
I thought protected classifies as BC break too, because class open to
extension will break if being extended and signature changes between
upgrades.

вт, 18 дек. 2018 г., 2:39 Michał Bundyra notifications@github.com:

webimpress:
@alextech As far as I know (if I am wrong please someone correct me) - in ZF rules for public interface are "public" methods and "protected" only when these are described as can be used in extension (for example in the documentation). I don't remember where I read about it, I can't find it now :(

src/FileInput.php
svycka: sorry I mean `protected` :) and I thought that if you set method `protected` it's like saying that you are safe to extend it if you need and we will not change it.

weierophinney:
@svycka @alextech and @webimpress —

Our policy since the ZF1 days has been to treat only `public` as part of the public API for BC purposes. We tend to scrutinize removals of `protected` methods a little closer, but, as @webimpress notes, unless it was documented as part of an internal API that users can rely on (e.g., how symfony/console documents `configure` and `execute`), we don't treat changes as BC issues.

In this particular case, there's little reason for extension, and the new changes make creating new strategies for file uploads easier.

svycka:
@weierophinney I am ok with this change just maybe this should be noted in changelog.
And good to know that protected can be removed or changed any time, will be more careful using them with ZF components

Slamdunk:
Like https://github.com/zendframework/zend-filter/pull/70 this PR introduced a BC Break. Hope to isolate the issue and post a PR soon.

PR #175, opened 2018-12-21
Fix regression introduced in 2.9.0 when overriding FileUpload default validator

Slamdunk:
https://github.com/zendframework/zend-inputfilter/pull/172 introduced a BC Break: till release-2.8.3 the default validator was set as alias `'fileuploadfile'`. In a functional test environment it must be overridden, and thus until version 3 the alias used must be kept `'fileuploadfile'`.

https://github.com/zendframework/zend-inputfilter/blob/799ad48ed1666d3c62126fec73dd20453b3a9e4d/src/FileInput.php#L190-L192

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,14 +22,15 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
-- Nothing.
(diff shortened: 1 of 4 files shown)


weierophinney:
Thanks, @SlamDunk!

PR #180, opened 2019-04-10
Add validator's priority retrieval into Factory

mtagliab:
I noticed that it was impossible to set validator's priority using the "array notation", even if this feature was already there as it already exists for filters (see line 384 of Factory.php).

To add the priority, it will be enough to add the `priority` key in the array:
```php
$inputFilter[] = [
    'name' => 'text_elt',
    'required' => true,
    'filters' => [
        ['name' => 'StripTags'],
        ['name' => 'StringTrim']
    ],
    'validators' => [
        [
            'name' => 'StringLength',
            'options' => [
                'max' => 256
            ],
            'priority' => 2
        ]
    ]
];
```

### FactoryTest fix
I also noticed that the FactoryTest class had a minor issue, which was that validators where not tested inside the `testFactoryWillCreateInputWithSuggestedValidators` method. 
I fixed the loop and added a final assert to verify that all the validators have been tested.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -425,7 +425,11 @@ class Factory
                 if (isset($validator['break_chain_on_failure'])) {
                     $breakChainOnFailure = $validator['break_chain_on_failure'];
                 }
-                $chain->attachByName($name, $options, $breakChainOnFailure);
+                $priority = ValidatorChain::DEFAULT_PRIORITY;
+                if (isset($validator['priority'])) {
+                    $priority = $validator['priority'];
+                }
+                $chain->attachByName($name, $options, $breakChainOnFailure, $priority);
                 continue;
             }
 
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -425,10 +425,7 @@ class Factory
                 if (isset($validator['break_chain_on_failure'])) {
                     $breakChainOnFailure = $validator['break_chain_on_failure'];
                 }
-                $priority = ValidatorChain::DEFAULT_PRIORITY;
-                if (isset($validator['priority'])) {
-                    $priority = $validator['priority'];
-                }
+                $priority = isset($validator['priority']) ? $validator['priority'] : ValidatorChain::DEFAULT_PRIORITY;
                 $chain->attachByName($name, $options, $breakChainOnFailure, $priority);
                 continue;
             }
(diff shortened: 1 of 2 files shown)


froschdesign:
@webimpress 
I think an unit test is missing here. Like this one:

https://github.com/zendframework/zend-inputfilter/blob/a23e67d41de09bbc325543acb756ba448f182c20/test/FactoryTest.php#L685-L726

webimpress:
@froschdesign test added. Would you mind to have a look on it, please?

It's a bit tricky as I mentioned in the commit comment - we can't use `getValidators` on `ValidatorChain`, as the result is not sorted by priority. When validators are called (in `isValid`) these are ordered, and this is how I check it.

froschdesign:
@webimpress 
> When validators are called (in `isValid`) these are ordered, and this is how I check it.

👍

src/Factory.php
```diff
@@ -425,10 +425,7 @@ protected function populateValidators(ValidatorChain $chain, $validators)
                 if (isset($validator['break_chain_on_failure'])) {
                     $breakChainOnFailure = $validator['break_chain_on_failure'];
                 }
-                $priority = ValidatorChain::DEFAULT_PRIORITY;
-                if (isset($validator['priority'])) {
-                    $priority = $validator['priority'];
-                }
+                $priority = isset($validator['priority']) ? $validator['priority'] : ValidatorChain::DEFAULT_PRIORITY;
```
mtagliab: Even if this is the most logic way to do it, I used the "ugly" format to maintain coherence with the upper statement (`breakChainOnFailure`).
If you want to use the ternary operator, it would be nicer to re-format the previous check too, so there aren't different coding styles in the same file.

webimpress: @mtagliab 
I've changed it to match the logic from `populateFilters`:
https://github.com/zendframework/zend-inputfilter/blob/c29cc31ba7e30da1ed64851d82387eefc7fe9a25/src/Factory.php#L384

As you can see the coding style is mixed in the file...

webimpress:
Thanks, @mtagliab!

src/Factory.php
mtagliab: Actually, you are right, I didn't see that.

PR #181, opened 2019-05-09
Fixes #177 - MVC InputFilterPluginManager missing InputFilterAbstractFactory config

### LINKED ISSUE #177: MVC InputFilterPluginManager missing InputFilterAbstractFactory config
InputFilterPluginManager in MVC environment does not create input filter defined in config.

### Code to reproduce the issue

```php
$config = [
    'input_filter_specs' => [
        MyInput::class => [
            //...
        ],
    ],
];

$inputFilter = $container->get(InputFilter\InputFilterPluginManager::class)->get(MyInput::class);

```

### Expected results

InputFilterPluginManager creates defined InputFilter

### Actual results

Exception thrown 

> Plugin of type MyInput is invalid; must implement Zend\InputFilter\InputFilterInterface or Zend\InputFilter\InputInterface

Imho this happens because Module.php provide incomplete config.

```
    public function getConfig()
    {
        $provider = new ConfigProvider();

        return [
            'service_manager' => $provider->getDependencyConfig(),
            'input_filters' => $provider->getInputFilterConfig(), // this line is missing
        ];
    }
```

froschdesign:
Related to #177 

/cc @vaclavvanik

### DIFF
--- a/test/ModuleTest.php
+++ b/test/ModuleTest.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * @link      https://github.com/zendframework/zend-inputfilter for the canonical source repository
+ * @copyright Copyright (c) 2019 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://framework.zend.com/license New BSD License
+ */
+
+namespace ZendTest\InputFilter;
+
+use Interop\Container\ContainerInterface;
+use PHPUnit\Framework\TestCase;
+use Zend\InputFilter\InputFilterAbstractServiceFactory;
+use Zend\InputFilter\InputFilterPluginManager;
+use Zend\InputFilter\InputFilterPluginManagerFactory;
+use Zend\InputFilter\Module;
+
+class ModuleTest extends TestCase
+{
+    /** @var Module */
+    private $module;
+
+    public function setUp()
+    {
+        $this->module = new Module();
+    }
+
+    public function testGetConfigMethodShouldReturnExpectedKeys()
+    {
+        $config = $this->module->getConfig();
+
+        $this->assertInternalType('array', $config);
+
+        // Service manager
+        $this->assertArrayHasKey('service_manager', $config);
+
+        // Input filters
+        $this->assertArrayHasKey('input_filters', $config);
+    }
+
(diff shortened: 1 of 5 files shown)


test/ModuleTest.php
```diff
@@ -0,0 +1,99 @@
+<?php
+/**
+ * @link      https://github.com/zendframework/zend-inputfilter for the canonical source repository
+ * @copyright Copyright (c) 2019 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://framework.zend.com/license New BSD License
```
webimpress:
```
* @license   https://github.com/zendframework/zend-inputfilter/blob/master/LICENSE.md
```

please

The same in all other added files.

Ref: https://github.com/zendframework/maintainers/blob/master/template/src/ConfigProvider.php#L5

froschdesign: _Should be automatically covered by the coding standard. At the moment there is already a green check mark._

webimpress: Yeh, hopefully in v2 of ZCS.

vaclavvanik:
@froschdesign Good job +1

test/ModuleTest.php
```diff
@@ -0,0 +1,99 @@
+<?php
+/**
+ * @link      https://github.com/zendframework/zend-inputfilter for the canonical source repository
```
Xerkus: Similar change, should be `@see` instead of `@link`
https://github.com/zendframework/maintainers/blob/master/template/src/ConfigProvider.php#L3

test/ModuleTest.php
```diff
@@ -0,0 +1,99 @@
...
+        // Input filters
+        $this->assertSame(
+            [
+                'abstract_factories' => [
+                    InputFilterAbstractServiceFactory::class,
+                ],
+            ],
+            $config['input_filters']
+        );
+    }
```
Xerkus: I am not particularly fond of testing config provider by repeating config. This test will break with any change to config.

I would split this test into two. One asserting that certain factories and aliases for main container are set (not exact array matching), so that adding more factories won't make test fail. And second asserting that abstract service factory is registered for input filter manager.

webimpress:
Thanks, @froschdesign!

